PR #5, opened 2015-10-30
Redux + Basic Edit Tag Screen

clloyd:
Adds a few utilties
 - `.eslintrc` file (`npm i` to fetch eslint and react plugin) for javascript code consistency.
 - `redux (`jspm install` to get deps) for frontend data flow

Edit Screen
 - 'Update' actions triggered on every key stroke
 - 'Save' action triggers server side save Save button only shown when the


jamespamplin:
:+1: good work!

PR #7, opened 2015-10-30
Webpack Builds + Janus credentials support

clloyd:
Webpack for the JS build pipeline

### Production builds 
 - Gone from 440KB -> ~200KB
 - No need for SystemJS ~50KB
 - 2 Javascript requests now 1.

### Development Builds
 - 430 Request for javascript per page load is now 1.
 - 2-3s wait for JS after page load is replaced with 0.5s compile on file change.

### Hot Reloading Mode!
 - Changes in react components immediately trigger the component to be updated in an open browser (Magic!)

Things to note:
 - JSPM has been completely removed, so no need to `jspm install` all packages are managed through npm.

## Setup

New nginx config needs to be applied in the usual manner, will also require a run of `setup.sh`.

There are also two new ways to start the app...

1. `./scripts/start.sh` will start sbt and the webpack watcher. Changing javascript requires a refresh to see. This is very close to the live implementation.
2. `./scripts/client-dev.sh` will start sbt and the webpack-dev-server (now with added hot reloading!). This enabled hot reloading (check your chrome console for connection status)

It's of course still possible to use `sbt run` and compile the latest production version of the js assets with `./setup.sh`


hmgibson23:
Very nice!

clloyd:
@uplne This is the webpack build pipeline we chatted about last week.

jamespamplin:
:+1: great work!

PR #8, opened 2015-11-02
Convert Sections to be stored in reflux store.

clloyd:
We require sections in the path construction so moved fetch out of SectionSelect Component and into the general store, prevents multiple fetches of the same data.

(Also added a guardian favicon to remove error in console)


public/components/utils/SectionSelect.react.js
```diff
@@ -1,38 +1,30 @@
...
-            type: 'json'
-        }).then(function(resp) {
-            self.setState({sections: resp});
-        }).fail(function(err, msg){
-            console.log('failed', err, msg);
-        });
-    }
+  constructor(props) {
+    super(props);
+    this.state = {sections: []};
```
steppenwells: If i understand the rest of this the empty sections in state is no longer used.

clloyd: yep you're correct! Will push a change

steppenwells:
Looks good :+1:

PR #11, opened 2015-11-02
Adds legally sensitive flag

steppenwells:


clloyd:
:+1: Brill! Will add UI once merged.

PR #12, opened 2015-11-03
Adds BasicTagInfo Component for Name

clloyd:
This adds a BasicTagInfo Component responsible for updating the various names a tag can have. 

screen shot 2015-11-03 at 12 00 15

This borrows ui from composer for locking and unlocking related pieces of information.


steppenwells:
Looks ok, I'll admit some of the svg lost me.

:+1:

PR #13, opened 2015-11-03
Polyfill Object.assign 

clloyd:
Updated babel to version 6 and include object.assign transform.


steppenwells:
My objects are now assigning
My webs are being packed.

:+1:

PR #14, opened 2015-11-05
Sw sequences

steppenwells:
Adds sequences for tagId and sectionId

These should be accessed as:

```scala
Sequences.tagId.getNextId
```


clloyd:
:+1:

PR #15, opened 2015-11-05
Create New Tags

clloyd:
This adds a new client side endpoint `tag/create` and allows POST requests to `/api/tag`

Creation is handles through a `Tag/Create` component (The existing `Tag` component is now `Tag/Display`) which constructs a blank tag in it's own state, then passes that to the existing TagEdit components. Nothing is communicated to the server until the save button is clicked.

This also includes the basics for different Form Elements based on Tag Type, e.g. section is only displayed when 'topic' is selected as tag type.


public/components/Tag/Create.react.js
```diff
@@ -0,0 +1,99 @@
...
+import TypeSelect from '../utils/TypeSelect.react';
+import SaveButton from '../utils/SaveButton.react';
+import {creatableTags} from '../../constants/tagTypes.js';
+
+const BLANK_TAG = {
+  externalReferences: [],
+  hidden: false,
+  legallySensitive: false,
+  pageId: 123456789,
+  path: '/this/should/be/set/serverside',
```
clloyd: path and pageId are to be removed

public/components/Tag/TagEdit.react.js
```diff
@@ -14,23 +19,37 @@ export default class TagEdit extends React.Component {
       }));
     }
 
+    //This will contain the logic for different tag forms (keyword vs contributor etc...)
```
clloyd: This will probably end up being separate components for each type before it gets unreasonable

public/components/Tag/formComponents/TagName.react.js
```diff
@@ -46,6 +46,12 @@ export default class TagBasicInfo extends React.Component {
   }
 
   onUpdateSlug(e) {
+
+    //Prevent updating slug when locked
+    if (!this.props.unlockSlug) {
+      return;
```
clloyd: Slug is locked by default.

public/components/Tag/TagEdit.react.js
```diff
@@ -14,23 +19,37 @@ export default class TagEdit extends React.Component {
...
+        return false;
+      }
+
+      if (this.props.tag.type === tagTypes.topic) {
+        return (<div className="tag-edit__input-group">
+          <label className="tag-edit__input-group__header">Section</label>
+          <SectionSelect selectedId={this.props.tag.section} sections={this.props.sections} onChange={this.onUpdateSection.bind(this)}/>
+        </div>);
+      }
+    }
```
steppenwells: Technically section is not a topic specific field, but happy to leave it here now as an example of different fields for types (and given we've only got topics currently)

public/components/utils/TypeSelect.react.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+export default class SectionSelect extends React.Component {
+
```
steppenwells: TypeSelect

public/constants/tagTypes.js
```diff
@@ -0,0 +1,5 @@
+export const topic = 'topic';
+export const series = 'series';
+export const contributor = 'contributor';
+
+export const creatableTags = [topic, series, contributor];
```
steppenwells: We might want to push the type list server side at some point, ok here for now.

steppenwells:
couple of minor comments but otherwise :+1:

PR #16, opened 2015-11-05
Add description field

clloyd:
Adds description field to TagEdit component

Requires #15


steppenwells:
Looks like that would indeed edit the description :+1:

PR #17, opened 2015-11-05
Clientside Validation

clloyd:
Adds validation utility, which when you pass a tag will return an array of errors. 

This does add `ReactCSSTransitionGroup` so will require an `npm i` in public

Hide the save button and show some validation messages: 

screen shot 2015-11-05 at 16 13 53


steppenwells:
Looks good so far. I'm going to have a couple of questions when I add path validations...

:+1:

PR #18, opened 2015-11-05
Improved Save Buttons Design

clloyd:
It's the same Save/Reset button we've grown to love, with a fresh new design.

screen shot 2015-11-05 at 17 07 18

SaveButton now accepts a isHidden prop, which if true will hide the buttons. Previously this was being handled by both parent components, so moved inside.


steppenwells:
All the magic appears to be in the styles, I guess it does what it says it does...

:+1:

clloyd:
It's really the addition of ReactCSSTransitionGroup with is essentially a react version of ng-animate

PR #19, opened 2015-11-05
TypeSelect shown but disabled on display screen.

clloyd:
This adds the TypeSelect screen in a disabled state to the Display screen, so you can see the tag type!

screen shot 2015-11-05 at 17 28 01


steppenwells:
:+1:, but we might want to revisit how we show the type in future. It feels more important than just a disabled dropdown. That said I'm not the UX expert (uxpert) so lets see what the users / AJ think

clloyd:
Definitely agree!

PR #20, opened 2015-11-06
Bring sass under webpack, move package.json to project root.

clloyd:
Sass is now compiled by webpack, giving us hot-reloading in client-dev mode.

This also moves the package.json (and node_modules) out of /public and into the root, this reduces the time taken to refresh a page after an asset file change from around 5 seconds to 1 second, and builds from around 11 mins to 5 mins.

# !!!

After merge you'll need to delete the node_modules folder inside your public directory and re-run `./scripts/setup.sh`


jamespamplin:
Fantastic! great stuff :+1:

clloyd:
@steppenwells I'll merge this, but as noted will allow you to delete node modules in your public folder (and greatly speed up sbt startup speeds!)

PR #22, opened 2015-11-09
Reduce NPM errors and set NODE_ENV=production on circle Builds

clloyd:
This separates the dependencies required for the production build, from the dependencies required only for the development builds (these are now devDependencies, not installed by circle)

Will reduce the amount of packages being installed by 4 per build, and remove the socket.io install errors which aren't required for the production builds.


public/devserver.js
```diff
@@ -11,29 +11,31 @@ var wpConfig         = require('./build_config/webpack.devserver.conf.js');
...
+  stats:   {
+    colors: true
+  }
 });
 
 //
 // Exports
 //
 
+//wpServer.use('/public', express.static('public'));
```
steppenwells: We probably don't need this line...

steppenwells:
:+1:

PR #25, opened 2015-11-10
Client side CAPI Stats

clloyd:
screen shot 2015-11-10 at 12 58 54

Component that takes a tag and collects CAPI usage stats.

This contains a constant with the CAPI URL, my next PR will extract that into a configuration object in state.

### DIFF
--- a/public/util/capi.js
+++ b/public/util/capi.js
@@ -1,0 +1,16 @@
+import Reqwest from 'reqwest';
+
+const CAPI_BASE = 'https://internal.content.guardianapis.com';
+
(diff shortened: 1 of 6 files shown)


public/util/capi.js
```diff
@@ -0,0 +1,16 @@
+import Reqwest from 'reqwest';
+
+const CAPI_BASE = 'https://internal.content.guardianapis.com';
```
clloyd: Will be extracted into a configuration object server side and stored in redux.

PR #26, opened 2015-11-10
Fix Path Locking for exiting tags.

clloyd:
Fixxed some cases around name locking could cause path to be changed on an existing tag


steppenwells:
looks good :+1:

PR #27, opened 2015-11-10
Sw path validation

steppenwells:
Adds path validation to check if the path is already in use.


public/components/Tag/Create.react.js
```diff
@@ -35,10 +37,52 @@ class TagCreate extends React.Component {
...
+        }
+      } else {
+        if(current.section) {
+          return true;
+        } else {
+          return false;
+        }
+      }
+    }
+
```
clloyd: Could this be:

```
if (!updated.section) {
  return !!current.section
}

return updated.section !== current.section

```

clloyd:
:+1: With suggestion for those if else if else blocks.

public/components/Tag/Create.react.js
steppenwells: or even just return `updated.section !== current.section;`

PR #28, opened 2015-11-11
CAPI Stats component

clloyd:
Added capi stats 
screen shot 2015-11-11 at 15 33 50

This adds a CapiClient util, which needs to be called with a path and key, then the returned object be used like any other api util (see CapiStats component)

This also adds a config pull from S3.


steppenwells:
Looks good :+1: etc.

PR #29, opened 2015-11-11
Adds "Category" to Topic Tags

clloyd:
As discussed in the meeting this adds the ability to select a category for a topic tag. 

The category is optional in the Tag model (as existing tags don't have it) but is a mandatory field in the interface.

### DIFF
--- a/app/model/Tag.scala
+++ b/app/model/Tag.scala
@@ -18,7 +18,7 @@ case class Tag(
   hidden: Boolean = false,
   legallySensitive: Boolean = false,
   comparableValue: String,
-  category: Option[String],
+  categories: Set[String] = Set(),
   section: Option[Long],
   description: Option[String] = None,
   parents: Set[Long] = Set(),
@@ -41,7 +41,7 @@ object Tag {
       (JsPath \ "hidden").format[Boolean] and
       (JsPath \ "legallySensitive").format[Boolean] and
       (JsPath \ "comparableValue").format[String] and
-      (JsPath \ "category").formatNullable[String] and
+      (JsPath \ "categories").formatNullable[Set[String]].inmap[Set[String]](_.getOrElse(Set()), Some(_)) and
       (JsPath \ "section").formatNullable[Long] and
       (JsPath \ "description").formatNullable[String] and
       (JsPath \ "parents").formatNullable[Set[Long]].inmap[Set[Long]](_.getOrElse(Set()), Some(_)) and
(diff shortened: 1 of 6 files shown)


steppenwells:
I think tags can have multiple categories. The British Library is both a place and an organisation.

clloyd:
@steppenwells Can now have multiple categories.

theefer:
Semi-philosophical question: does applying a tag always mean applying both
categories of the tag?

Eg

Double-murder at the British Library
George Osborne cuts all funding to the British Library

On Thu, 12 Nov 2015, 10:34 Christopher Lloyd <notifications@github.com>
wrote:

> @steppenwells <https://github.com/steppenwells> Can now have multiple
> categories.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guardian/tagmanager/pull/29#issuecomment-156065236>.
>

-- 

------------------------------
This e-mail and all attachments are confidential and may also be 
privileged. If you are not the named recipient, please notify the sender 
and delete the e-mail and all attachments immediately. Do not disclose the 
contents to another person. You may not use the information for any 
purpose, or store, or copy, it in any way.  Guardian News & Media Limited 
is not liable for any computer viruses or other material transmitted with 
or as part of this e-mail. You should employ virus checking software.
 
Guardian News & Media Limited is a member of Guardian Media Group plc. Registered 
Office: PO Box 68164, Kings Place, 90 York Way, London, N1P 2AP.  Registered 
in England Number 908396

clloyd:
@theefer I would say yes, but to me categories describe the tag not the content it's attached to. I suppose if if the categories were so different... (although I struggle to think of an example) Then two tags could be created? It's a really interesting issue. 

Perhaps it might be worth discouraging multiple categories (while still allowing it)

theefer:
Proper ontologies would have different tags for different categories of a concept. If this is just to retrofit categories onto the existing navigational tags we have, then I guess it's ok to have multiple categories per tag, as long as we understand and recognize that it is very limiting and just a short-term hack prior to migrating to a more usefully modelled ontology.

steppenwells:
:+1: on the code.

We'll have a much bigger discussion of how categories will be used in the future, Nathan is beginning to talk to frontend and work out how we might use categories in the tools. I fully expect the model will change between here and the final implementation.

As for single tags per category I think this will come down to editorial discretion, we already separate the Microsoft (business) and Microsoft (technology) tags as there is enough content to make this viable, for the British Library less so (if such a tag even exists). For some hypothetical future ontological system it might be we'll use a compound to tag + category construct.

theefer:
OK, thanks for the detailed answer!

PR #30, opened 2015-11-11
Sw thrift model

steppenwells:
Adds thrift model for tags and sections.

Raises tag creations as TagEvents to a kinesis stream.

cc @mchv  - Does this model work from CAPI's point of view?


clloyd:
:+1: Makes sense, once merged can look at using those ENUMs to replace the client side constants.

thrift/tag.thrift
```diff
@@ -0,0 +1,68 @@
...
+    5: required string internalName;
+
+    /** the publicly displayed name for the tag */
+    6: required string externalName;
+
+    /** the tag owned url slug, the path is derived from this, the section and the tag type */
+    7: required string slug;
+
+    /** the tag is not displayed on the site if hidden is true */
+    8: required bool hidden;
```
mchv: I don't think this is indexed currently and used anymore, but anyway I think we should avoid visual directive.

thrift/tagEvent.thrift
```diff
@@ -0,0 +1,20 @@
+include "tag.thrift"
+
+namespace scala com.gu.tagmanagement
+
+enum EventType {
+    CREATE = 0,
```
mchv: Recently we decided to not add a `CREATE` type for both atoms and content as in practice, updates and creation are handled the same way.

thrift/tag.thrift
```diff
@@ -0,0 +1,68 @@
...
+
+    /** legally sensitive tags surpress content showing related content and showing up in related content results */
+    9: required bool legallySensitive;
+
+    /** The natural sort key used for ordering tag lists, puts surname first, removes 'the' etc. */
+    10: required string comparableValue;
```
mchv: I don't want to repeat the last time drama :smile: but I don't think putting `surname` first is a good example of natural sorting.

mchv:
Others thing to add in the future are `Contributor` metadata:
 * `firstName`
 * `lastName`
 * `twitterHandle`
 * `emailAddress`) 

and `Podcast` metadata:
 * `linkUrl`
 * `copyright`
 * `author`
 * `itunesBlock`
 * `explicit`
 *  `itunesSubscriptionUrl`

PR #31, opened 2015-11-12
Initialises log shipping as a module

steppenwells:
Rather than as a side effect of constructing the App controller


clloyd:
:+1:

PR #33, opened 2015-11-13
Create.js Now uses redux store for it's state

clloyd:
Bumped React version (`npm i` please)
Redux'd Create.js, now behaves alot like Display.js
Fixed issues with UI state persisting between different newly created tags i.e. `<textbox>`'s are a pain


steppenwells:
:+1:

PR #34, opened 2015-11-13
Initial Batch Tagging UI

clloyd:
This is an initial stab at the Batch tagging UI, covering the searching and selection of content pieces.

screen shot 2015-11-13 at 15 01 30

This doesn't include filtering, or doing anything useful with the selected pieces of content.


.tern-project
```diff
@@ -0,0 +1,16 @@
+{
```
clloyd: This is a config for [tern](https://github.com/ternjs/tern) which makes auto suggestion for JS a bit smarter, trying it out.

steppenwells:
:+1: Lets get this merged and I'll slap a backend on it.

PR #35, opened 2015-11-13
Exclude moment locale info

clloyd:
One line change that cuts 125KB minified from production build. We're not using moment's locale information so it can be excluded.


steppenwells:
whatevs :+1:

PR #36, opened 2015-11-16
Sw tag event stream

steppenwells:
This PR keeps the internal tag lookup cache up to date by subscribing to the tag updates kinesis stream.

So as not to create a potentially infinite number of kinesis tracking dynamo tables the app grabs a node id at startup to use to configure the kinesis app name. This node id is returned to the pool for reuse when the app terminates.

The update tag api endpoint has been reworked to update the kinesis, otherwise all the other code would be pointless.


app/model/command/UpdateTagCommand.scala
```diff
@@ -0,0 +1,20 @@
...
+import play.api.Logger
+import repositories.TagRepository
+import services.KinesisStreams
+
+
+case class UpdateTagCommand(tag: Tag) extends Command[Tag] {
+  override def process: Option[Tag] = {
+    Logger.info(s"updating tag ${tag.id}")
+
+    val result = TagRepository.upsertTag(tag)
```
clloyd: Does this mean the node gets the update twice? once here and once when it appears on the stream?

steppenwells: No, I removed the local insert from the repo, all updates come via the stream.

clloyd:
:+1:

PR #39, opened 2015-11-25
Series Tag Model

clloyd:
This adds series support to the tag model (Which is really just a podcastMetadata section)

This adds definitions for 
 - PodcastMetadata
 - Image
 - ImageAsset

### DIFF
--- a/thrift/image.thrift
+++ b/thrift/image.thrift
@@ -1,0 +1,29 @@
+namespace scala com.gu.tagmanagement
+
+struct ImageAsset {
+
+    /** the url of the asset */
+    1: required string imageUrl;
+
+    /** the width of the asset */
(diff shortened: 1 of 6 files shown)


thrift/image.thrift
```diff
@@ -0,0 +1,29 @@
...
+    /** the url of the asset */
+    1: required string imageUrl;
+
+    /** the width of the asset */
+    2: required i64 width;
+
+    /** the height of the asset */
+    3: required i64 height;
+
+    /** the height of the asset */
```
steppenwells: this is not the height.

steppenwells:
rename content type to mimeType as thats what we're storing.

steppenwells:
Other than minor comments looks good.

PR #41, opened 2015-11-27
Podcast Support

clloyd:
This adds podcast editing support to the interface when a series tag is being edited/created.

Also cleaned up the styling of all the checkbox based edit fields


steppenwells:
Looks good :+1:

PR #42, opened 2015-11-27
Sw perform batch tag

steppenwells:
Lots here and it's not fully tested yet (the flex-feeds / R2 batch tag sides are not complete yet). But here it is for reference. Includes:

 * job running module
 * basic job ui component on the tag screen
 * raises batch tagging operations
 * tagging operation thrift model (the model also supports merge and reindex operations.)


clloyd:
:+1: merge conficts :(

PR #43, opened 2015-11-27
Remove old capiclient init from CapiStats

clloyd:
Quick fix to CapiStats to no longer init the Capi client


steppenwells:
Guess this is OK. I'll admit I've not done the archeology to find how the client is now inited, but I'm taking it on trust it's working

:+1:

PR #44, opened 2015-12-04
Podcast Images (and Image Metadata)

clloyd:
This adds the non-GRID image add/removal functionality for Podcast Image (in the form of a general component for use everywhere)

On inputting of a recognised url (jpg or png on https://static.guim.co.uk) the component will query the `/support/image-metadata` to recieve:

``` JSON

{
   "width": 720,
   "height": 600,
   "size": 121116,
   "id": "23ea8766375c48498c1fabc1ee448af1a2f9cb69",
   "mimeType": "image/png"
}
```

This can then be used to update the podcast metadata

### DIFF
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -1,0 +1,22 @@
+package controllers
+
+import play.api.libs.json.Json
+import play.api.mvc.{Action, Controller}
+import play.api.Logger
+import services.ImageMetadataService
+
+object Support extends Controller with PanDomainAuthActions {
+
+  def imageMetadata(imageUrl: String = "") = AuthAction {
+
+    val imageMetadata = ImageMetadataService.fetch(imageUrl)
+    Ok(Json.toJson(imageMetadata))
+
+    ImageMetadataService.fetch(imageUrl).map { imageMetadata =>
+      Ok(Json.toJson(imageMetadata))
+    }.getOrElse(NotFound)
+
+  }
+
+
+}
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -2,21 +2,13 @@ package controllers
 
 import play.api.libs.json.Json
 import play.api.mvc.{Action, Controller}
-import play.api.Logger
 import services.ImageMetadataService
 
 object Support extends Controller with PanDomainAuthActions {
 
-  def imageMetadata(imageUrl: String = "") = AuthAction {
-
-    val imageMetadata = ImageMetadataService.fetch(imageUrl)
-    Ok(Json.toJson(imageMetadata))
-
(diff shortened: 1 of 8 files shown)


app/controllers/Support.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import play.api.libs.json.Json
+import play.api.mvc.{Action, Controller}
+import play.api.Logger
+import services.ImageMetadataService
+
+object Support extends Controller with PanDomainAuthActions {
+
+  def imageMetadata(imageUrl: String = "") = AuthAction {
+
```
steppenwells: this should probably be an ApiAuthAction so it 419's rather than redirects to google

app/controllers/Support.scala
```diff
@@ -0,0 +1,22 @@
...
+    val imageMetadata = ImageMetadataService.fetch(imageUrl)
+    Ok(Json.toJson(imageMetadata))
+
+    ImageMetadataService.fetch(imageUrl).map { imageMetadata =>
+      Ok(Json.toJson(imageMetadata))
+    }.getOrElse(NotFound)
```
steppenwells: This appears to be implemented twice.

public/components/TagEdit/formComponents/TagImageEdit.react.js
```diff
@@ -0,0 +1,152 @@
...
+    });
+  }
+
+  renderAddButton() {
+    if (this.state.metadataFetchStatus === FETCH_STATES.fetching) {
+      return (<div className="tag-edit__image__add">Checking Url...</div>);
+    }
+
+    if (this.state.metadataFetchStatus === FETCH_STATES.error) {
+      return (<div className="tag-edit__image__add--error"><i className="i-cross-red" />Url Not Valid</div>);
```
steppenwells: Maybe add a bit more explanation about what constitutes a valid image url, that it needs to be on static.guim... etc.

app/services/ImageMetadata.scala
```diff
@@ -0,0 +1,67 @@
...
+
+        Some(ImageMetadata(image.getWidth, image.getHeight, bytes.size, mediaId, mimeType))
+      }
+      case c => {
+        Logger.warn(s"failed to get image metadata for $uri. response code $c, message ${response.body}")
+        throw ImageMetadataFetchFail(s"failed to get image metadata for $uri. response code $c, message ${response.body}")
+      }
+    }
+  } catch {
+    case NonFatal(e) => None
```
steppenwells: The try / catch here will swallow the ImageMetadataFetchFail exception and result in a 404 being returned to the frontend. It might be better to remove the try catch here and handle the exceptions in to controller (or just let them propagate as a 503 )

steppenwells:
a few small comments but otherwise looks good. :+1:

PR #45, opened 2015-12-04
Sw audit log

steppenwells:
Audit log for tag activity. Logging works fine, but the display is still pretty crufty:

screen shot 2015-12-04 at 17 54 32

Still to do, I'll get on to this when I get back from holiday, unless someone is feeling charitable in the meantime:

 * [x] convert those millis into human friendly dates (moment)
 * [x] is there a better way to invoke the render methods, surely I'm using self too much here
 * [x] reverse list so most recent first, also possibly truncate.
 * [x] refresh component on save (without this the latest record is missing after update)
 * [x] tidy the display, the table is not really working is it?
 * [ ] Add reports to status page using the audit by operation endpoint


steppenwells:
This should now be good to merge. we'll do the status page stuff later.

clloyd:
@steppenwells fixed the broken build (was old cached deps on circleci)

PR #47, opened 2015-12-07
Contributor Tag Support

clloyd:
Slight refactor of TagEdit to clean up switching between different tag types.

Support for contributor type tags.


rrees:
This looks fine but are you sure you want to model the Contributor data like-for-like with R2?

You could create a Contributor Metadata sub-model like the Podcast information instead.

clloyd:
@rrees Yeah that would be much cleaner, will change it to that format this morning

PR #48, opened 2015-12-08
Modeled Client Settings, Pass TagType constants

clloyd:
This adds a ClientConfig model that can be expanded with further server side constants, removing duplication between the server-side and client-side constants. 

TagType throughout the application is now been powered by the thrift definition Enum.


app/model/ClientConfig.scala
```diff
@@ -0,0 +1,15 @@
+package model
+
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{JsValue, Json, JsPath, Format}
+
+case class ClientConfig(capiUrl: String, capiKey: String, tagTypes: List[String])
+
+object ClientConfig {
+
+  implicit val clientConfigFormat: Format[ClientConfig] = (
```
rrees: I don't really get what this is doing, or alternatively why it is being done this way.

clloyd: The object? or the implicit val? 

Trying to find the cleanest way to JSON encode this information, which is going to expand as we pass more info to client on load.

steppenwells:
Looks good, loving removing the blog type

:+1:

PR #51, opened 2015-12-10
Add a wrapper around Scribe for use with React

clloyd:
This adds ReactScribe, a wrapper around a scribe instance, turning into a React Component. Surfacing the regular `value` and `onChange` that you would expect with any other input.

screen shot 2015-12-10 at 15 01 42


rrees:
:+1: good job we made that possible before we actually needed the functionality

PR #55, opened 2015-12-16
Permissions!

clloyd:
This adds support for `permissions.gutools.co.uk` and limits batch_tag to allowed users only.

### DIFF
--- a/app/controllers/App.scala
+++ b/app/controllers/App.scala
@@ -7,11 +7,15 @@ import repositories.{TagSearchCriteria, TagRepository}
 import play.api.Logger
 import services.Config
 import com.gu.tagmanagement.{TagType}
+import permissions._
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
 
 
 object App extends Controller with PanDomainAuthActions {
 
-  def index(id: String = "") = AuthAction {
+  def index(id: String = "") = AuthAction.async { req =>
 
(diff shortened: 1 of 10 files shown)


public/routes/routes.js
```diff
@@ -1,6 +1,25 @@
 import React        from 'react';
 import { Route, IndexRoute } from 'react-router';
 
+<<<<<<< HEAD
```
theefer: Oops?

app/controllers/App.scala
```diff
@@ -20,9 +24,13 @@ object App extends Controller with PanDomainAuthActions {
       case None => routes.Assets.versioned(jsFileName).toString
     }
 
-    val clientConfig = ClientConfig(Config().capiUrl, Config().capiKey, TagType.list.map(_.name))
+    Permissions.getPermissionsForUser(req.user.email).map { permissions =>
```
theefer: Apologies if this is a stupid question because I don't know the semantics of the Permissions library, but what happens to this is the permissions service is down?

clloyd: Permissions are stored in S3, so downtime should be minimal, they are also cached by the application so only needs to get it successfully once, and if all else fails, falls back to the default, which is in this case... no access.

theefer: OK fair enough! :+1:

PR #57, opened 2016-01-05
Section Editing And Editions

clloyd:
This adds sections editing and editions to the tagmanager.

Sections can be viewed at https://tagmanager.local.dev-gutools.co.uk/section and can be edited at https://tagmanager.local.dev-gutools.co.uk/section/<section id> 

Currently only the section name can be edited. This like the tags can then be saved or reset, and have full tag-like auditing.

screen shot 2016-01-05 at 15 34 12

Any international editions are shown along with paths, these can be delete or added to, each of these operations contacting a new endpoint directly, so pathmanager updates can be handled (without diff-ing editions on each update)

screen shot 2016-01-05 at 15 33 56

### DIFF
--- a/app/controllers/TagManagementApi.scala
+++ b/app/controllers/TagManagementApi.scala
@@ -1,7 +1,7 @@
 package controllers
 
 import model.command.CommandError._
-import model.command.{BatchTagCommand, PathUsageCheck, CreateTagCommand, UpdateTagCommand}
+import model.command._
 import model.jobs.{BatchTagAddCompleteCheck, Job}
 import org.joda.time.DateTime
 import permissions.Permissions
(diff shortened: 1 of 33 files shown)


app/controllers/TagManagementApi.scala
```diff
@@ -86,15 +86,38 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
+    }.getOrElse {
+      BadRequest("Expecting Json data")
+    }
+  }
+
+  def addEditionToSection(id: Long) = APIAuthAction { req =>
+    req.body.asJson.map { json =>
+      val editionName = (json \ "editionName").as[String]
 
+      try {
```
hmgibson23: Might be easier to use the scala Try for this and the above ... should be a bit simpler to read.

hmgibson23:
Looks good :+1:

PR #58, opened 2016-01-05
XML Dump of all tags for inCopy to use

hmgibson23:
This adds a route that adds a dump of all the tags in XML format. I haven't done the from time method yet as that will require a bit more work and we wanted to find out how much it was used (\cc @blishen)

### DIFF
--- a/app/controllers/TagManagementApi.scala
+++ b/app/controllers/TagManagementApi.scala
@@ -1,5 +1,4 @@
 package controllers
-
 import model.command.CommandError._
 import model.command.{BatchTagCommand, PathUsageCheck, CreateTagCommand, UpdateTagCommand}
 import model.jobs.{BatchTagAddCompleteCheck, Job}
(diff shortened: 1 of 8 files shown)


app/controllers/TagManagementApi.scala
```diff
@@ -140,4 +141,11 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
     Ok(Json.toJson(jobs))
   }
 
+  def getTagsAsXml() = APIAuthAction.async {
+    Future(TagRepository.loadAllTags.seq.map(_.asXml)) map { tags =>
```
clloyd: We have a cached version of this which would be alot faster. I don't think there's currently a method to access it though.

hmgibson23: Okay cool, will make a method for it.

app/controllers/TagManagementApi.scala
```diff
@@ -140,4 +140,11 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
     Ok(Json.toJson(jobs))
   }
 
+  def getTagsAsXml() = APIAuthAction.async {
```
steppenwells: We almost certainly don't want this endpoint to authenticated. The incopy processes that consume it will not be using panda (the current "security" mechanism that incopy uses is ip and user agent based).

It might be worth moving this endpoint to a new unsecured read only controller, separating out the public read api from the tag manager specific read write api.

hmgibson23: Okay. Will do

steppenwells:
Looks good, we should check that inCopy can understand the output, but :+1: for now

PR #59, opened 2016-01-05
Sw merge tag

steppenwells:
Adds api endpoints and jobs for tag merge and delete. I've not added the UI changes to trigger deletes and merges here but think it's probably a good idea to get this code out to the team before I head off to the states. the endpoints should probably be permissioned as well...

flexible feeds changes to support merging are here: https://github.com/guardian/flexible-feeds/pull/7


clloyd:
:+1:

PR #60, opened 2016-01-07
Hypermedia API for integrating with Flexible Content

hmgibson23:
This adds a Hypermedia API that essentially just mimics the existing tag api. This is so that we can make minimal changes to flex. I've had to make two additional classes ```TagEntity``` and ```SectionEntity``` so that the API returns the tags in the format flex expects. 

We should test this a lot but it should be as simple as dropping the root hyper url into flex and watching it work.


clloyd:
:+1: Wow!

theefer:
Cool :-)

PR #62, opened 2016-01-08
Added compression byte to all kinesis streams

### LINKED ISSUE #61: Inital Compression Byte missing for Porter
@oilnam  [1:10 PM] 
so we have a small issue

​[1:10] 
we have a custom byte at the beginning of the thrift data that tells us if and what compression we’re using

​[1:12] 
https://github.com/guardian/content-api/blob/master/porter/src/main/scala/com.gu.contentapi.porter/integration/ThriftDeserializer.scala#L30

​[1:13] 
so for consistency it would be good if you could add the `0` bit at the beginning of the thrift data

​[1:13] 
zero meaning “no compression"

clloyd:
This adds a compression byte (currently set to 0x00) to all kinesis streams so CAPI can consume. Also adds ability for consumer to strip the compression byte

Fixes #61

### DIFF
--- a/app/services/AWS.scala
+++ b/app/services/AWS.scala
@@ -64,7 +64,7 @@ object SQS {
   lazy val jobQueue = new SQSQueue(Config().jobQueueName)
 }
 
-class KinesisStreamProducer(streamName: String) {
+class KinesisStreamProducer(streamName: String, requireCompressionByte: Boolean = false) {
 
   def publishUpdate(key: String, data: String) {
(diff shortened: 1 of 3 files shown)


currysoup:
Out of interest, will we ever not need the compression flag?

clloyd:
It's my understanding that it's a CAPI specific thing, so wanted the flexibility to disable on future kinesis feeds to non-capi projects (I've no idea if that'll be a thing...).

I originally intended to not use it on our internal streams, but then we use some kinesis streams internally and for CAPI, so just turned on for all feeds.

Could default to true rather than false if that makes more sense.

PR #63, opened 2016-01-08
Moving thrift definitons to guardian/tags-thrift-schema

currysoup:
Moved the thrift schema into a separate repository and hosted on Bintray's JCenter. This means `scrooge-sbt-plugin` is no longer required since this project no longer builds the thrift.

`scrooge-core` is still required as a dependency for serialization.

This project now imports the thrift definitions as a libraryDependency.

The namespace of these definitions is still `com.gu.tagmanager` which can be changed if there's any desire for that.


hmgibson23:
:+1:

PR #65, opened 2016-01-11
Removing thrift folder, now in guardian/tags-thrift-schema

currysoup:


currysoup:
No longer needed due to a9655aa

PR #66, opened 2016-01-12
Made permissions more generic and setup permissions for endpoints

currysoup:
Moving from a granular permissions system (e.g. having a permission for `batch_tag`) to a more generic system with two levels: `admin` and `super-admin`. As such, an update to the `permissions` service is also required. Pull request submitted [here](https://github.com/guardian/permissions/pull/38).

### Server 
I've setup the API endpoints to perform a permissions check in a similar fashion to the `batch_tag` permissions check. This results in several Scala objects, each describing the permissions check for a given function; as can be seen [here](https://github.com/guardian/tagmanager/blob/sc-admin-permissions/app/permissions/PermissionActionCheck.scala#L24-L53). This allows each permissions check to have its own `restrictedAction` string which is logged if the user fails the permissions check. It should be possible to rewrite how these permissions are checked to remove the objects which would mean the model of composing ActionFilters might change - I'm happy to look into this if there's a desire to eliminate this ~30 LoC.

The following is a table of the various endpoints in the TagManagementAPI and their given permissions level.

| Func                     | Level       |
|--------------------------|-------------|
| addEditionToSection      | Super-Admin |
| removeEditionFromSection | Super-Admin |
| updateSection            | Super-Admin |
| mergeTag                 | Admin       |
| batchTag                 | Admin       |
| deleteTag                | Admin       |
| getSection               | Common      |
| updateTag                | Common      |
| getTag                   | Common      |
| createTag                | Common      |
| searchTags               | Common      |
| listReferenceTypes       | Common      |
| listSections             | Common      |
| checkPathInUse           | Common      |
| getAuditForTag           | Common      |
| getAuditForOperation     | Common      |
| getJobs                  | Common      |

### Client
I've also updated the client code to forward the user onto the unauthorized section if they don't have the correct permissions.


clloyd:
:+1:

PR #67, opened 2016-01-12
Add Github SSH keys and PathManager Timeout

clloyd:
This adds 
- @philmcmahon's github team ssh access :fireworks: 
- A much shorter connect timeout on the three requests made to pathmanager (previously they would timeout after 60s)
- Bumps AMI to latest


hmgibson23:
:+1:

PR #69, opened 2016-01-13
Update the XML export to match what comes out of R2

hmgibson23:
This changes the XML exported from tags and references to match what comes out of R2. This is needed so that inCopy (and Pluto?) can consume the right feed. 

I've also added the routes so that you can get a single tag. I will add the existing routes later.

\cc @clloyd


clloyd:
:+1:

PR #71, opened 2016-01-14
All XML routes in existing R2 index

hmgibson23:
This finalises all the routes in the current index. Still need to discuss the ones that currently 404 in code and see how they are used. 

The routes for fetching XML info about the tags are:
```
/tags/export/all
/tags/export/single/:id
/tags/export/merges/:since
```


hmgibson23:
ping @clloyd

clloyd:
@hmgibson23 do we pass that since in R2?, or do we need to track the last request time.

hmgibson23:
No it gets passed in and I haven't implement those routes yet anyway (doing them now)

clloyd:
:+1:

PR #73, opened 2016-01-18
Reindex endpoint, and auditing

currysoup:
Created a preliminary reindexing endpoint with auditing. This may be changed depending on the desires of the CAPI team with regards to reindexing.

### DIFF
--- a/app/controllers/Reindex.scala
+++ b/app/controllers/Reindex.scala
@@ -1,0 +1,24 @@
+package controllers
+
+import model.command.CommandError._
+import model.command.ReindexCommand
+import permissions.ReindexPermissionsCheck
+import play.api.libs.json._
+import play.api.Logger
+import play.api.mvc.Controller
+
+object Reindex extends Controller with PanDomainAuthActions {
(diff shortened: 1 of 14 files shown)


app/controllers/Reindex.scala
```diff
@@ -0,0 +1,24 @@
...
+import play.api.libs.json._
+import play.api.Logger
+import play.api.mvc.Controller
+
+object Reindex extends Controller with PanDomainAuthActions {
+  def reindex = (APIAuthAction andThen ReindexPermissionsCheck) { req =>
+    // Get the reindex id provided by CAPI
+    req.body.asJson.map { json =>
+        try {
+          ReindexCommand((json \ "id").as[String]).process.map{t => NoContent } getOrElse NotFound // TODO not found doenst make much sense here
```
hmgibson23: You could do an ```InternalServerError``` and put some values in it. Or another of the many play errors instead of NotFound.

currysoup:
Good catch on the TODO, thanks

hmgibson23:
:+1: Nice

PR #74, opened 2016-01-18
Tag with no content fix + UI Improvements

clloyd:
Fix for delete entering loop when no content returned from CAPI.

This fixes a few UI issues: 

 - `<TagSelect />` now shows the tag type,
 - `<TagSelect />` is now a sensible width
 - `<TagSelect />` suggestions should now appear above other ui elements,
 - `<TagNaming />` should now guess slugs much more accurately,
 - Tag types should now be referred to with a nicer name (Spaces not just camel case)


hmgibson23:
:+1:

PR #75, opened 2016-01-19
Add the final routes for Newspaper integration

hmgibson23:
This adds the last remaining routes for newspaper integration. 

The routes are:
```
http://r2.code.dev-gutools.co.uk/tools/newspaperintegration/tags/new/:since
http://r2.code.dev-gutools.co.uk/tools/newspaperintegration/tags/modified/:since
http://r2.code.dev-gutools.co.uk/tools/newspaperintegration/tags/deleted/:since
```


app/controllers/ReadOnly.scala
```diff
@@ -23,13 +24,50 @@ object ReadOnlyApi extends Controller {
...
-      {merges.map( x => x.asExportedXml)}
+    Ok(
+      <merges>
+        {merges.map( x => x.asExportedXml)}
       </merges>
     )
   }
+
+  def deletesAsXml(since: Long) = Action {
+    val deletes = JobRepository.getDeletes.map { job =>
```
clloyd: This isn't a record of all Jobs that have happened in the past, they are deleted once complete (the only reason they're there now is due to other issues with merge/delete)

Tag Audit is a better source of this information

hmgibson23: Okay cool - will update

app/controllers/TagManagementApi.scala
```diff
@@ -161,7 +161,7 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
     }
   }
 
-  def deleteTag(id: Long)= (APIAuthAction andThen DeleteTagPermissionsCheck) { req =>
```
clloyd: Why are we removing this permission check?

conf/routes
```diff
@@ -51,10 +51,17 @@ GET        /hyper/tags                    controllers.HyperMediaApi.tags
 GET        /hyper/tags/:id                controllers.HyperMediaApi.tag(id: Long)
 OPTIONS    /*all                          controllers.HyperMediaApi.preflight(all: String)
 
+# Tag export for inCopy
 GET        /tags/export/all               controllers.ReadOnlyApi.getTagsAsXml
 GET        /tags/export/single/:id        controllers.ReadOnlyApi.tagAsXml(id: Long)
 GET        /tags/export/merges/:since     controllers.ReadOnlyApi.mergesAsXml(since: Long)
 
+# More newspaper integration
+GET       /tools/newspaperintegration/tags/modified/:since          controllers.ReadOnlyApi.modifiedAsXml(since: Long)
```
clloyd: We are going to have to change these urls in the consumers, so if we want something less ugly :D

hmgibson23: I guess so. I imagine for now they'll just want to switch over the hosts though

hmgibson23:
@clloyd - mind taking a look?

clloyd:
:+1:

PR #76, opened 2016-01-19
Handle 404 returned from CAPI when checking for deleted tags

clloyd:
This should allow delete jobs to complete, the CAPI response was returning an exception, rather than None when a tag could not be found (e.g. when checking if a tag has been deleted)


hmgibson23:
:+1:

PR #78, opened 2016-01-21
Microsites & Sections

clloyd:
- Bump thrift library
- Support Microsites
- Add Blog Tag Type
- Add new Section fields (isMicrosite and Discriminator)


hmgibson23:
:+1:

PR #79, opened 2016-01-21
Tag Level Permissions Checking 

currysoup:
Introducing permissions based on a tag's type. Previous permissions checking was on a tagmanager function level (batch tag, merge tag, etc.)

The server introduces a new ActionFilter which will check the request body for a JSON "type" field and use that fields value in a permissions lookup.

Client changes are basically representing these permissions to the user in an expected way. Previously editable text fields are now greyed and read-only. There's also a big orange warning bar telling them they don't have permission to edit the tag.

More implementation details are available in the commit logs.


public/actions/TagActions/getTag.js
```diff
@@ -11,11 +15,12 @@ function requestTagGet() {
...
 
-function recieveTagGet(tag) {
+function recieveTagGet(tag, canEdit) {
     return {
-        type:       TAG_GET_RECEIVE,
-        tag:        tag,
-        receivedAt: Date.now()
+        type:        TAG_GET_RECEIVE,
+        tag:         tag,
+        tagEditable: canEdit,
```
clloyd: Is tagEditable used here?

public/actions/TagActions/getTag.js
```diff
@@ -32,7 +37,22 @@ export function getTag(id) {
...
         return tagManagerApi.getTag(id)
-            .then(res => dispatch(recieveTagGet(res)))
+            .then(res => {
+
+              const store = getStore();
+              var permitted = store.getState().config.permittedTagTypes;
+              var tagType = res.type;
+
+              if (!permitted.some((e, i, a) => e == tagType)) {
+                dispatch(showWarning('You do not have permission to edit this tag type'));
```
clloyd: Is this the error that is being shown inside Display.js? I wouldn't keep this in the store. This should show the error in ReactApp (but don't use it for this situation)

public/components/Tag/Display.react.js
```diff
@@ -64,6 +64,41 @@ class TagDisplay extends React.Component {
...
+
+      if (permitted.some((e, i, a) => e == this.props.tag.type)) {
+        return <SaveButton isHidden={!this.isTagDirty() || !this.isTagValid()}
+                           onSaveClick={this.saveTag.bind(this)}
+                           onResetClick={this.resetTag.bind(this)}/>
+      }
+    }
+
+    renderWarningBar() {
+      if (!this.props.warning) {
```
clloyd: I think this should be specific to tagEditable    `if (!this.props.tagEditable)` not show the system wide warning

public/actions/TagActions/getTag.js
clloyd: I can't see it in the reducer

currysoup: Yeah, it's passed into the reducer and makes sure all the UI components are usable or disabled depending on permissions settings.

clloyd: Ok, missed it

clloyd:
:+1: Apart from the small comment

public/components/Tag/Display.react.js
currysoup: Good catch, this was written before the tagEditable property was implemented.

currysoup: I'll pull out the entire warning system and make the warning bar specific to permissions issues. If we need more warnings in the future we can generalise then. :tada:

PR #80, opened 2016-01-21
Fix some issues with merging

hmgibson23:
This adds a new step that will upload the merge operation to the audit table as previously we were just tracking  the old tag being deleted and we need to keep track of the whole merge.


clloyd:
Great Catch! :+1:

PR #83, opened 2016-01-22
Made code environment

currysoup:
Just updated the Config.scala to point at the new tables, streams, and queue.


clloyd:
:+1: LGTM!

PR #84, opened 2016-01-22
Section Reindexing

currysoup:
Added section reindexing endpoint, command, job and step. 

Lots of renaming of the old command to reindexTag* 

New kinesis streams are in our AWS account.


clloyd:
:+1:

PR #86, opened 2016-01-25
PathManager using internal load balancer for CODE/PROD

clloyd:
Pathmanager is now accessing the internal load balancers directly, this config could be argued should be kept private so is being stored in S3 config.

Fixes tag creation on CODE


hmgibson23:
:+1:

PR #89, opened 2016-01-25
Update Discriminator on Section Edition Change

clloyd:
This adds logic to change the discriminator based on the current edition count, if 0 it should be "Navigation" else it should me "MultiEdition"


hmgibson23:
Looks good :+1:

PR #99, opened 2016-01-26
Removed categories component from view until after release

### LINKED ISSUE #94: Hide Categories
This will be enabled later, hide topic tag categories for the time being.

clloyd:
This comments out the TopicCategories select which won't be supported at launch (but added in soon after)

fixes #94


hmgibson23:
Might want to put a TODO or something to make it clear that at some point the functionality will be restored but :+1:

Issue #91, opened 2016-01-26
Remove Blog from creation list completely


clloyd:
No one should be able to create Blog tags

PR #102, opened 2016-01-26
Limits Blog Tag creation, show deprecation

clloyd:
Blog Tag is now deprecated (Shows deprecated in tag select) and requires super admin permissions - #91 

ContentType tag is now behind a super admin permission for creation


hmgibson23:
:+1:

PR #106, opened 2016-01-28
Reauth and save progress spinner

currysoup:
Automatic reauth when your cookies expire. The end behaviour is your log gets the 419s, there's a very slight delay and then your page loads as normal. I'm not 100% about the PandaReqwest promises, but it definitely seems to work.

Also bonus save progress spinner.


clloyd:
:+1:

PR #110, opened 2016-01-28
Improve JobStatus Table to provide better step information

clloyd:
Adds:
 - Ability to Delete Jobs (if permitted)
 - More step information (how many pieces of content still have old tag) etc...


app/permissions/PermissionActionCheck.scala
```diff
@@ -43,6 +43,11 @@ object DeleteTagPermissionsCheck extends PermissionActionFilter {
   val restrictedAction = "delete tag"
 }
 
+object DeleteJobPermissionsCheck extends PermissionActionFilter {
+  val testAccess: String => Future[PermissionAuthorisation] = Permissions.testUser(Permissions.TagSuperAdmin)
```
currysoup: Will be simpler to merge if you set the permission requirements to just admin

public/components/Status.react.js
```diff
@@ -21,46 +22,30 @@ export default class Status extends React.Component {
...
-                <th>Type</th>
-                <th>User</th>
-                <th>Progress</th>
-              </tr>
-            </thead>
-            <tbody className="status__results">
-              {this.state.jobStatus.sort((a, b) => a.id > b.id ? 1 : -1).map(this.renderListItem, this)}
-            </tbody>
-          </table>
+          <JobTable jobs={this.state.jobStatus} triggerRefresh={this.fetchJobs.bind(this)}  disableDelete={!this.props.config.permissions.tag_super_admin}/>
```
currysoup: Again here, super admin is going today so might be easier to deal with it before merge (especially since it's javascript you'll just end up with it being undefined :cry: )

public/util/tagManagerApi.js
```diff
@@ -212,5 +212,13 @@ export default {
       method: 'get',
       type: 'json'
     });
+  },
+
+  deleteJob: (id) => {
+    return Reqwest({
```
currysoup: During merge this should become a PandaReqwest :)

clloyd:
@currysoup Should be good now

app/controllers/TagManagementApi.scala
```diff
@@ -208,4 +208,10 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
     Ok(Json.toJson(jobs))
   }
 
+  def deleteJob(id: Long) = (APIAuthAction andThen DeleteJobPermissionsCheck) {
+    JobRepository.deleteJob(id)
```
currysoup: Do we want to deal with `Dynamo.jobTable.deleteItem` throwing exceptions? Or just let play framework deal with it as a 500?

clloyd: We're not dealing with it elsewhere, it's probably clearer as a 500 tbh

currysoup:
:+1: :rocket:

PR #111, opened 2016-01-28
Remove super admin permission

currysoup:
Anything that was previously super admin is now just standard admin,

Batch tag is now common (requires no special permissions).

Requires: https://github.com/guardian/permissions/pull/39


clloyd:
:+1:

Issue #116, opened 2016-01-29
Parents should only be shown for Topic Tags


clloyd:
Tag search responses

PR #123, opened 2016-02-01
Hide Tag Parents for non-topic tags

clloyd:


clloyd:
fixes #116

public/components/TagContext/TagRelationship/TagRelationship.react.js
```diff
@@ -100,6 +101,10 @@ export default class TagRelationship extends React.Component {
         return false;
       }
 
+      if (this.props.tag.type !== tagTypes.topic.name) {
```
currysoup: I've seen topics being mapped to keyword tags. Does this mean this rule should also compare to keyword tags?

clloyd: No, in this case it's fine, they're only being mapped to Keywords when they're crossing the boundaries to old systems.

currysoup:
:+1:

PR #126, opened 2016-02-01
Increasing contrast of save bar

currysoup:
If this isn't adequate please add a comment :smile: 

Previously:

previous

New:

current


clloyd:
Good idea, lets just do this and get some concrete feedback before doing more. I think once it's used it'll be obvious.

PR #127, opened 2016-02-01
Tag and Section XML output to match R2 more closely

clloyd:
This adds support for the Global section, fixes some Option.toString's and adds a section XML endpoint


currysoup:
:+1: :palm_tree: :tropical_drink:

PR #132, opened 2016-02-03
Updated Path Calculation for publication and content type

clloyd:
This updates the path calculation for publication and content type tags, previously we were calculating the capi id, but these tag types the id differs from the path, this calculates the correct path


currysoup:
:+1:

PR #133, opened 2016-02-03
CAPI preview support

clloyd:
This adds a proxy route which performs a request to preview CAPI. Updated the UI to reflect if an article is published or not

### DIFF
--- a/app/controllers/App.scala
+++ b/app/controllers/App.scala
@@ -41,7 +41,15 @@ object App extends Controller with PanDomainAuthActions {
         }
       }
 
-      val clientConfig = ClientConfig(Config().capiUrl, Config().capiKey, allTags, permittedTags.toList, permissions, "/reauth")
+      val clientConfig = ClientConfig(
+        capiUrl = Config().capiUrl,
+        capiPreviewUrl = "/support/previewCapi",
+        capiKey = Config().capiKey,
+        tagTypes = allTags,
+        permittedTagTypes = permittedTags.toList,
+        permissions = permissions,
+        reauthUrl = "/reauth"
+      )
 
       Ok(views.html.Application.app("Tag Manager", jsLocation, Json.toJson(clientConfig).toString()))
     }
--- a/app/services/Config.scala
+++ b/app/services/Config.scala
@@ -27,6 +27,9 @@ sealed trait Config {
 
   def capiUrl: String = getRequiredStringProperty("capi.url")
   def capiKey: String = getRequiredStringProperty("capi.key")
+  def capiPreviewUrl: String = getRequiredStringProperty("capi.preview.url")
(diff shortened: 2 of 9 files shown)


app/services/Config.scala
```diff
@@ -65,7 +68,6 @@ sealed trait Config {
   }
 
   lazy val permissionsStage: String = readTag("Stage") match {
-    case Some("DEV") =>  "CODE"
```
currysoup: Is there a dev permissions stage?

LATaylor-guardian: On CAPI there is only CODE or PROD, no DEV.

clloyd: You need to spin up a local instance of permissions, which seems like overkill

clloyd:
@LATaylor-guardian Is this enough protection on top of preview? It's behind panda auth

app/services/Config.scala
currysoup: Does this need to be addressed before merge? Unsure of the implications.

clloyd: So all this does is map the DEV stage of TagManager (Local) to the CODE stage of permissions. It's prefereable to do this, or we'd have to all run permissions locally to change local permissions

currysoup: But this line is being removed? Won't this mean if someone wants to run in DEV they will have to run a permissions locally?

Sorry, I'm a bit confused :cry:

clloyd: My mistake, will re-add

LATaylor-guardian:
@clloyd I'll take a look over this in a bit.

app/controllers/App.scala
```diff
@@ -41,7 +41,15 @@ object App extends Controller with PanDomainAuthActions {
         }
       }
 
-      val clientConfig = ClientConfig(Config().capiUrl, Config().capiKey, allTags, permittedTags.toList, permissions, "/reauth")
+      val clientConfig = ClientConfig(
+        capiUrl = Config().capiUrl,
+        capiPreviewUrl = "/support/previewCapi",
+        capiKey = Config().capiKey,
```
LATaylor-guardian: Why do you need to return the capiUrl and capiKey to the client if you are going via the proxy?

clloyd: This is because we are only using preview for the batch tagger, and for live CAPI we are going direct from client (To fetch tag usage stats etc)

LATaylor-guardian: So just to verify, you will never expose the preview URL for CAPI on the client side?

clloyd: never, we expose the proxy route which requires panda and the live (non-internal) capi url

LATaylor-guardian: Okay brilliant :)

LATaylor-guardian:
This looks okay to me from a CAPI security point of view. :+1:

PR #136, opened 2016-02-04
Mapping Manager Fix

clloyd:
Small fix to mapping manager to only show mappings that match the currently selected mappings, previously would add all mappings of tags that had one matching mapping. mapping...


currysoup:
:earth_africa: :mag: :+1:

PR #137, opened 2016-02-04
Added paging to batchtagging

currysoup:
# :book: Batch Tag Paging :book: 

Added pagination to batch tagging!

pagination

## Current status
* Pages are currently 50 content items long, this can be changed trivially but 50 seems like a reasonable number and prevents React.js from slowing down too much.
* Pages are loaded lazily then cached.
* When a user selects a tag and changes page, their tag is still selected. Their content item will still be selected when they navigate back to the original page.
* The PageNavigator component is not coupled to tagmanager at all. It can be reused fairly easily.
* Currently shows 5 pages to the left and to the right of the current page, in addition to the first page so the user can easily jump back.

### DIFF
--- a/public/actions/CapiActions/searchCapi.js
+++ b/public/actions/CapiActions/searchCapi.js
@@ -1,10 +1,18 @@
 import debounce from 'lodash.debounce';
 import {searchContent} from '../../util/capiClient';
 
+export const CAPI_CLEAR_PAGES = 'CAPI_CLEAR_PAGES';
 export const CAPI_SEARCH_REQUEST = 'CAPI_SEARCH_REQUEST';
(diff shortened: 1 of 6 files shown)


public/actions/CapiActions/searchCapi.js
```diff
@@ -48,6 +66,18 @@ export function searchCapi(searchString, params) {
     };
 }
 
+export function switchPage(page) {
+    return dispatch => {
```
clloyd: This could just be 
``` JS
export function switchPage(page) {
    return {
        type: CAPI_SWITCH_PAGE,
        page: page
    }
}
```

clloyd:
:+1:

PR #140, opened 2016-02-05
Added Created XML route.

clloyd:


currysoup:
I'm getting a 301 from the endpoint when using http and a 504 gateway timeout when using https.

Is there anything special you need to do to use this PR?

clloyd:
504 is because local nginx config is timing out too quickly (this thing takes AGES).

currysoup:
:+1:

PR #143, opened 2016-02-08
Added clearErrorBar to routes

currysoup:
Can be used to clear errors when moving from one page to another.

If we ever need to do more than one thing on a onLeave event we can
use something like ramda to compose functions.


clloyd:
:+1:

PR #145, opened 2016-02-08
DO NOT MERGE BEFORE OR ON MONDAY ReadOnly endpoint Performance and Section Lookup Cache

currysoup:
Times gone from over 15 minutes down to about 20 seconds in initial testing.


clloyd:
Can you mark this DO NO MERGE in title?

currysoup:
Some approximate speedups:

| Endpoint | Speedup |
|--------------|--------------|
| /tags/export/all | x30 |
| /sections/export/all | x2 |
| /tools/newspaperintegration/created/0 | x3*|
| /tools/newspaperintegration/modified/0 | x7 |

* Rerunning this resulted in a speedup of about x30

clloyd:
:+1:

currysoup:
Seeing some weird behaviour with the kinesis stream clients not picking up data. Perhaps related to the execution contexts again. Investigating now...

currysoup:
Just going to close this off. We currently scan the entire section DB into memory when performing read only options. This is plenty fast enough (since there are so few sections) and doesn't require us to handle cache coherency which is neat.

PR #149, opened 2016-02-09
Added permissions workaround

currysoup:
Teeny tiny play module to preload the permissions as part of start up.


lindseydew:
What's the timing bug with the permissions client? Is it to do with populating the cache on start up? If this is the case I would highly recommend defaulting to contacting the web service to find out a users permission. The service should be the source of truth for ascertaining a users permission, and I think it could be confusing to have a client that overwrites this. I may have misunderstood this PR and happy to discuss

clloyd:
:+1:

PR #152, opened 2016-02-10
Tracking Tags

clloyd:
This adds support for tracking tags (although not selectable in the UI)

This also updates the path calculation logic to include the more difficult tracking tag & will only show the tag edit fields once a type is selected, resetting the tag when it is changed

### DIFF
--- a/public/components/Tag/Create.react.js
+++ b/public/components/Tag/Create.react.js
@@ -70,7 +70,7 @@ class TagCreate extends React.Component {
     }
 
     checkPathInUse(tag) {
-      tagManagerApi.checkPathInUse(tag.type, tag.slug, tag.section)
+      tagManagerApi.checkPathInUse(tag.type, tag.slug, tag.section, tag.trackingInformation ? tag.trackingInformation : undefined)
         .then(res => this.setState({pathInUse: res.inUse}))
         .fail((error) => {
           console.log(error);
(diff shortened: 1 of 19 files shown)


public/components/Tag/Create.react.js
```diff
@@ -96,8 +96,8 @@ class TagCreate extends React.Component {
                 <label className="tag-edit__input-group__header">Tag Type</label>
                 <TypeSelect selectedType={this.props.tag.type} types={this.props.config.permittedTagTypes} onChange={this.onUpdateType.bind(this)}/>
               </div>
-              <TagEdit tag={this.props.tag} sections={this.props.sections} updateTag={this.updateTag.bind(this)} tagEditable={this.props.tagEditable}/>
-              <TagValidationErrors validations={this.generateValidationErrors()} />
+              {this.props.tag.type ? <TagEdit tag={this.props.tag} sections={this.props.sections} updateTag={this.updateTag.bind(this)} tagEditable={this.props.tagEditable}/> : false}
+              {this.props.tag.type ? <TagValidationErrors validations={this.generateValidationErrors()} /> : false }
```
currysoup: When would a tag not have a type?

currysoup:
A small question but otherwise :+1:

public/components/Tag/Create.react.js
clloyd: When it's not been selected yet, brand new tag

PR #156, opened 2016-02-11
Clicking elements in a list no longer replaces the current history

currysoup:
Preventing the Tag and Section lists from writing over history, which lead to some perhaps surprising behaviour.

The create tag/section actions still replace the current history. It seems to make sense since you're not undoing the creation action when pressing back.


clloyd:
:+1:

PR #158, opened 2016-02-11
Basic Auth Added to prevent access before launch

clloyd:
Note: This is not meant to be secure private information! It's just meant to stop people accessing the system before launch.


theefer:
Isn't there a Play utility that let's you do this in a single line of code?

currysoup:
:+1:

clloyd:
@theefer I couldn't find one that was smaller than this, but if you can point me in the right direction can use in the future!

PR #159, opened 2016-02-12
TagEntity updated to match composer representation

clloyd:
Fixes an issue in the hypermedia api where topics should be rewritten as Keywords, and external reference values is actually "token"


hmgibson23:
:+1:

Issue #146, opened 2016-02-09
Front End Cleanup


clloyd:
Description/Scribe
 - [x] Formatting not functional (was previously...??) Chrome DEV bug, fixed in canary

Date Picker
 - [x] Should we be using that 3rd Party component? (#166)
 - [x] Need to add icons, fix styling (currently copied and pasted styling) (#166)

Publication Tag https://tagmanager.code.dev-gutools.co.uk/tag/2
 - [ ] Publication information remove buttons are unstyled
 - [ ] Newspaper Book Sections only show id's should fetch name similar to parents (#161)
 - [ ] Newspaper Book Sections are unstyled, maybe a box similar to parents? (#187)

Tag Context
 - [x] Add new parent relationship is unclear/ugly, could use improvement (#178)
 - [ ] Add new reference is Un-Styled

Image Uploader
 - [ ] Separating @steppenwells's upload module into a styled module (ideally with upload feedback etc)

Tag Audit
 - [x] Add the ability to see a summary of the tags involved, these are stored with the audit record, ideally shouldn't be visible constantly, maybe clicking on a row shows more info? (#171)

Lists
- [ ] List views have no hover / CTA indication - e.g. [the tag list](https://tagmanager.code.dev-gutools.co.uk/)

PR #161, opened 2016-02-15
Feature/157 Newspaper Books name is displayed

### LINKED ISSUE #157: Create CS API to fetch multiple tags at once and use it to fetch  Newspaper Book Section tags
Related to #146

thekarel:
For #146: Newspaper Book Sections only show id's should fetch name similar to parents 

Fixes #157


clloyd:
I'm a bit worried about this, hooking into TAG_GET_RECEIVE seems a little scary when this only affects 2 tags out of 50,000.  It seems really "high touch"

PR #166, opened 2016-02-15
Feature/162 clean up date picker widget

### LINKED ISSUE #162: Clean up date picker widget
Re Date Picker in #146 

The widget seems to be fine, but needs to be cleaned up a bit.

thekarel:
* Calendar looks and acts as the original

Fix #162


clloyd:
What effect does this have on the size of the application?

thekarel:
It should be the same as before the PR, as the whole CSS was copy pasted AFAIK.

clloyd:
:+1: was worried we were including new assets

PR #170, opened 2016-02-16
Batchtag fixes

clloyd:
- Can now search without first typing a query (e.g. just based on filter)
 - ContentType path calculations are now done before CAPI request (Content type is an exception where Capi ID is different from our Path)


currysoup:
:+1:

PR #173, opened 2016-02-16
Added a function to the tag manager api to clean up tags

currysoup:


public/util/tagManagerApi.js
```diff
@@ -57,9 +85,10 @@ export default {
   },
 
   createTag: (tag) => {
+    const clean = cleanTag(tag);
```
clloyd: Should this be done on update rather than create and save? So the client side is using the object that will be sent to the server?

currysoup: I did consider this but I thought it best to not change anything from under React since this isn't anything *wrong* per se with the React, just how the tagmanager server side processes things.

Would be trivial to move it into an action if you'd prefer.

clloyd: It's not really changing anything under react, react is just rendering the result of the clean data, which to me seems preferable? Also means we keep the idea of the state.tag is what we're sending back to the server, which makes debugging nicer.

currysoup: I'll move it into an action because I agree React should see exactly what we're sending.

Should it be in the updateTag action though? That would mean we need to make 100% sure we always call update prior to calling save. I'm not sure if I'm missing some logic here but that seems like a bit of a manual process.

Create and save both return the tag back to React so just calling cleanTag prior to dispatching the save request should have a similar effect without requiring the programmer ensure update is called prior to saving.

clloyd: Every update triggers updateTag updating this.state.tag, when you call save
or create that's passed this.state.tag so update will always have been
called.

On Tue, 16 Feb 2016 at 13:49 Sam Cutler <notifications@github.com> wrote:

> In public/util/tagManagerApi.js
> <https://github.com/guardian/tagmanager/pull/173#discussion_r53011599>:
>
> > @@ -57,9 +85,10 @@ export default {
> >    },
> >
> >    createTag: (tag) => {
> > +    const clean = cleanTag(tag);
>
> I'll move it into an action because I agree React should see exactly what
> we're sending.
>
> Should it be in the updateTag action though? That would mean we need to
> make 100% sure we always call update prior to calling save. I'm not sure if
> I'm missing some logic here but that seems like a bit of a manual process.
>
> Create and save both return the tag back to React so just calling cleanTag
> prior to dispatching the save request should have a similar effect without
> requiring the programmer ensure update is called prior to saving.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guardian/tagmanager/pull/173/files#r53011599>.
>

-- 

------------------------------
This e-mail and all attachments are confidential and may also be 
privileged. If you are not the named recipient, please notify the sender 
and delete the e-mail and all attachments immediately. Do not disclose the 
contents to another person. You may not use the information for any 
purpose, or store, or copy, it in any way.  Guardian News & Media Limited 
is not liable for any computer viruses or other material transmitted with 
or as part of this e-mail. You should employ virus checking software.
 
Guardian News & Media Limited is a member of Guardian Media Group plc. Registered 
Office: PO Box 68164, Kings Place, 90 York Way, London, N1P 2AP.  Registered 
in England Number 908396

currysoup:
Updated to move to updateTag action :smile:

clloyd:
:+1: I'd check this carefully on CODE though to make sure it doesn't affect anything obscure

PR #174, opened 2016-02-16
Strip styling fonts from ReactScribe

clloyd:
Add scribe sanitise with very strict rules to the ReactScribe component


currysoup:
post hoc :+1:

PR #176, opened 2016-02-16
Adds a "section tag id" on the section for flex tag proposer usage

clloyd:
This adds an extra section tag id field to the hypermedia api, this will remove a lot of back and forth currently done on the flex side.

Also bumps ssl ARN


rrees:
:+1:

PR #177, opened 2016-02-17
Add local composer to CORS-able domains in code

clloyd:
Local composer should use code tagmanager (so people don't need to run tagmanager locally), this adds local composer as a "cors"able domain in CODE tagmanager


currysoup:
:+1: 
:confetti_ball:

PR #178, opened 2016-02-17
Parents: better UX

thekarel:
Re #146 

* Add parent opens / closes add widget
* Add widget displayed next to add button, not over it
* Parent tag list presented in a table view
* Tag list above add button

parents


clloyd:
:+1:

PR #182, opened 2016-02-17
Fixing section audit username reporting

currysoup:
The issue was we had an implicit parameter `Option[User]` which defaulted to `None`. At some point in the past users stopped being passed around as `User` objects and were passed around as a `Option[String]`. This silently broke the audit creation since the lack of an implicit `Option[User]` caused the function to simply default to `None`.

Here's a screenshot of the new behaviour: 
sectionupdate

I've checked through and all the places where audits are constructed have `Some(username)`

### DIFF
--- a/app/model/SectionAudit.scala
+++ b/app/model/SectionAudit.scala
@@ -42,20 +42,20 @@ object SectionAudit {
     }
   }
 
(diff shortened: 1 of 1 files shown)


clloyd:
:+1:

PR #183, opened 2016-02-18
Updated Cloudformation to take SSL arn as a parameter

clloyd:
NOTE: the cloudformation will need to be applied manually before this is merged.

Also bumps AMI


currysoup:
:+1: :lock:

PR #185, opened 2016-02-18
Bugfix/ui update fixes

thekarel:
Allows Circle to build the JS bundle, also fixes a table view's styling

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "css-loader": "^0.22.0",
     "expose-loader": "^0.7.0",
     "extract-text-webpack-plugin": "^0.9.1",
+    "file-loader": "^0.8.5",
     "history": "1.17.0",
     "lodash.debounce": "^3.1.1",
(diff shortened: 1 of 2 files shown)


clloyd:
I think these shouldn't be devDependencies, we're currently using the dev/non-dev dependencies to determine what needs to be installed by circle and what doesn't. If those new dependencies are required to build the project I'd argue that they're not devDependencies. We want the builds to be as fast as possible.

thekarel:
I completely agree but that still wouldn't solve our problem with the CI server.

clloyd:
:+1:

clloyd:
@currysoup just noticed that repository being added, any reason for that?

thekarel:
To suppress an npm warning.

clloyd:
Ok, then we can use our actual repo address

thekarel:
Oh, I see what you mean. My mistake, sorry!

PR #181, opened 2016-02-17
Allowed images to be sent from s3-uploader

currysoup:
Images can now be added from S3 uploader:

A valid image:
imageupload

A "valid" URL which does no have an image at the end:
imageupload_notreal

PR #189, opened 2016-02-19
Adding more valid image extensions

currysoup:
S3 uploader doesn't change the extension file name so we've got to handle most common cases.

Added the very common `jpeg` and also the far less common `PNG`, `JPG` and `JPEG` I've seen some cameras give use.

Thanks to Akash for pointing this out in #181


clloyd:
:+1:

PR #190, opened 2016-02-23
Reindex Progress 

currysoup:
implemented as per the capi specification


app/controllers/Reindex.scala
```diff
@@ -7,43 +7,51 @@ import play.api.libs.functional.syntax._
...
-            Ok(Json.toJson(ExpectedItemCount(count)))
-          } getOrElse InternalServerError
-        } catch {
-          commandErrorAsResult
-        }
-      }.getOrElse {
-        BadRequest("Expecting reindex id in body")
+    try {
+      if (ReindexProgressRepository.isSectionReindexInProgress) {
+        Forbidden
```
clloyd: Just to confirm this is the response that CAPI is expecting in this case?

app/model/jobs/ReindexSteps.scala
```diff
@@ -1,44 +1,64 @@
...
+import scala.util.control.NonFatal
 
-// Tags
-
-case class ReindexTags(capiJobId: String) extends Step {
+case class ReindexTags() extends Step {
   override def process: Option[Step] = {
-    TagLookupCache.allTags.get.grouped(Config().reindexTagsBatchSize).foreach { tags =>
-      KinesisStreams.reindexTagsStream.publishUpdate(capiJobId, Tag.createReindexBatch(tags))
+    val total = TagLookupCache.allTags.get.size
```
clloyd: Would it be a good idea to have this coming from the Dynamo directly? don't want to propagate cache errors to CAPI?

app/controllers/Reindex.scala
LATaylor-guardian: 403 is what flexible content is returning so lets stick with that for consistency. 429 is a good shout though.. didn't know that one!

currysoup: 429 would perhaps be more appropriate indeed. Apparently CAPI is expecting a 403 if there's on already running. I'd be more than happy to change it if CAPI can accept 429 or something more meaningful than forbidden.

app/model/ReindexProgress.scala
```diff
@@ -0,0 +1,85 @@
+package model;
+
+import com.amazonaws.services.dynamodbv2.document.Item
+import org.cvogt.play.json.Jsonx
+import play.api.libs.json._
+import play.api.Logger
+import scala.util.control.NonFatal
+
+case class ReindexProgress(`type`: String, status: String, docsSent: Int, docsExpected: Int) {
```
LATaylor-guardian: docsSent should be documentsIndexed
docsExpected should be documentsExpected

app/model/ReindexProgress.scala
```diff
@@ -0,0 +1,85 @@
...
+
+object ReindexProgress {
+  val TagTypeName = "tag"
+  val SectionTypeName = "section"
+
+  val InProgress = "in progress"
+  val Failed = "failed"
+  val Completed = "completed"
+  val Cancelled = "cancelled"
+  val Unknown = "unknown"
```
LATaylor-guardian: You shouldn't need this one. This was an internal implementation detail on our side.

app/model/ReindexProgress.scala
```diff
@@ -0,0 +1,85 @@
...
+      throw e
+    }
+  }
+
+  def unknownTag() = {
+    ReindexProgress(TagTypeName, Unknown, 0, 0)
+  }
+
+  def unknownSection() = {
+    ReindexProgress(SectionTypeName, Unknown, 0, 0)
```
LATaylor-guardian: As above.

app/model/ReindexProgress.scala
```diff
@@ -0,0 +1,85 @@
...
+    Json.parse(item.toJSON).as[ReindexProgress]
+  } catch {
+    case NonFatal(e) => {
+      Logger.error(s"failed to load reindex progress ${item.toJSON}")
+      throw e
+    }
+  }
+
+  def unknownTag() = {
+    ReindexProgress(TagTypeName, Unknown, 0, 0)
```
LATaylor-guardian: As above.

app/model/jobs/ReindexSteps.scala
currysoup: I was thinking about this. If we don't trust our caches we might as well just scan out the entire dynamo table for every request where we use the cache...

Would make tagmanager pretty much stateless (?) which is nice.

clloyd: I know what you mean, and it's worth the debate, but I think for this it should be using the Dynamo directly even if we keep caches.

app/model/ReindexProgress.scala
currysoup: It's only used when the dynamo table we're storing our state in is empty for some reason. Either someone cleared it out or there has never been a reindex ran from that table.

clloyd:
:+1: with cache comment

PR #194, opened 2016-02-26
Added `path` to TagEntity (for hypermedia api)

clloyd:
This is so flex can store paths

### DIFF
--- a/app/model/TagEntity.scala
+++ b/app/model/TagEntity.scala
@@ -16,7 +16,8 @@ case class TagEntity(
   slug: String,
   section: SectionEntity,
   parents: Set[EmbeddedEntity[TagEntity]] = Set(),
-  references: List[ReferenceEntity] = Nil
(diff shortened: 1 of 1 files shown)


currysoup:
:+1: Seems gucci

clloyd:
:sunglasses:

app/model/TagEntity.scala
```diff
@@ -44,7 +45,8 @@ object TagEntity {
         tag.slug,
         getTagSection(tag.section),
         parents.map(x => EmbeddedEntity[TagEntity](HyperMediaHelpers.tagUri(x))),
-        tag.externalReferences.map(ReferenceEntity(_))
+        tag.externalReferences.map(ReferenceEntity(_)),
+        tag.path
       )
   }
```
currysoup: getTagSection could use a flatMap

PR #196, opened 2016-03-02
Add sponsorship support

steppenwells:
Adds sponsorship and paid content tag type plus edit and index screens. There's a module to do the automatic activation and expiry of sponsorships.

The builds should go green and spring into life once the new thrift definition is published:

https://github.com/guardian/tags-thrift-schema/pull/8

I've tested locally and it should be safe to pump the new thrift data at CAPI before support is added (that's how thrift works), although this should be verified on code.

Some of the css work could do with a bit of a once over but all screens are functionally complete, and getting this out will free up the CAPI work. Since the sponsorship routes are not linked to yet that tidy up can happen after this monster is in.


app/model/command/CreateSponsorshipCommand.scala
```diff
@@ -0,0 +1,72 @@
...
+      sponsorLogo = sponsorLogo,
+      sponsorLink = sponsorLink,
+      tag = tag,
+      section = section,
+      targeting = targeting
+    )
+
+    SponsorshipRepository.updateSponsorship(sponsorship).map { createdSponsorship =>
+
+      if(status == "active") {
```
clloyd: Do we need to maintain active and validFrom validTo, could this be within the valid times and not active? Save maintaining a computed state

steppenwells: The state is derived from the dates but the marker is needed so that the schedule launcher and expirer know which things to update.

public/components/SponsorshipSearch.react.js
```diff
@@ -0,0 +1,139 @@
+import React from 'react';
+import { Link } from 'react-router';
+import tagManagerApi from '../util/tagManagerApi';
+import SponsorshipList from './SponsorshipList/SponsorshipList.react';
+
+const searchStatuses = {
```
clloyd: Can we move these into a constants file (`js/constants`)

app/model/command/CreateSponsorshipCommand.scala
clloyd: Ok!

clloyd:
Looks good to me apart form very small JS comment @currysoup can you also take a look?

currysoup:
:+1:

PR #199, opened 2016-03-15
Microsite Unexpiry

clloyd:
This adds the ability for users with 'tag-admin' permissions to expire a Microsite, clicking the button will trigger a Preview CAPI request for all content within that section and then raises Kinesis messages directly to flexible-feeds with expired -> false.

Few other nice to haves added:
  - Dismiss "X" on the error bar to hide the error
  - Permissions Util for checking if the current user has a certain permission
  - Content API now support preview queries!


currysoup:
:+1: looks gucci

PR #200, opened 2016-03-16
TagList and SectionList are no longer Tables

clloyd:
This replaces the `<table>` with a div + flex based system. Allowing right click to open in new tab functionality. Applied to both the TagList and SectionList components.

### DIFF
--- a/public/components/SectionList/SectionList.react.js
+++ b/public/components/SectionList/SectionList.react.js
@@ -20,12 +20,20 @@ class SectionList extends React.Component {
     }
 
     renderListItem(section) {
+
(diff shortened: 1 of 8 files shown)


steppenwells:
The SponsorshipList was heavily cribbed from the tag and section lists so could probably do with a similar makeover.

Obviously this is a bit cheeky since the sponsorship stuff is not linked to yet so there's no reason for you to have remembered it. However it is merged so worth considering

clloyd:
@steppenwells updated to include SponsorshipList

steppenwells:
Looks good :+1:

PR #202, opened 2016-03-22
Fixing up a CAPI query size limitation

currysoup:
When a tag had a lot of associated content we failed to split up the content into small enough chunks leading to either the URL being too long or overrunning the max amount of content you can check at any one time.

This fix incrementally builds the tag count query and flushes the request if we overrun the limitations.

### DIFF
--- a/app/repositories/ContentAPI.scala
+++ b/app/repositories/ContentAPI.scala
@@ -22,19 +22,43 @@ object ContentAPI {
   private val apiClient = new LiveContentApiClass(Config().capiKey, Config().capiUrl)
   private val previewApiClient = new DraftContentApiClass(Config().capiKey, Config().capiPreviewUrl)
 
-  def countOccurencesOfTagInContents(contentIds: List[String], apiTagId: String): Int= {
+  def countOccurencesOfTagInContents(contentIds: List[String], apiTagId: String): Int = {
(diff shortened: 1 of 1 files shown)


app/repositories/ContentAPI.scala
```diff
@@ -22,19 +22,43 @@ object ContentAPI {
   private val apiClient = new LiveContentApiClass(Config().capiKey, Config().capiUrl)
   private val previewApiClient = new DraftContentApiClass(Config().capiKey, Config().capiPreviewUrl)
 
-  def countOccurencesOfTagInContents(contentIds: List[String], apiTagId: String): Int= {
+  def countOccurencesOfTagInContents(contentIds: List[String], apiTagId: String): Int = {
+    val builder = StringBuilder.newBuilder
+    var pageSize = 0
+
+    var total = 0
+
```
clloyd: Not sure if it's just me but this is quite hard to read, I think it's the countTags in 2 places.

Could you use the for loop to construct an array of strings ready for the various requests that you'll make, then can you could use a fold, making requests and combining count.

currysoup: You mean folding into a list of requests? Then processing all those lists after the fact?

clloyd: yeah, seems more functional?

PR #203, opened 2016-03-22
Add ability to exclude content based on tags

clloyd:
This adds the opposite of "Has Tags" will exclude content based on the tags specified.


currysoup:
Code looks good :dancers:

PR #205, opened 2016-03-23
Moving tagmanagers content API to preview

currysoup:
Moving all of tag managers content API requests to preview CAPI. Previously operations such as deleting a tag would only check content in live CAPI. Leaving dangling tag references in preview CAPI.


clloyd:
:+1: :+1: :+1: :+1: :+1: :+1: :+1:

Issue #206, opened 2016-03-29
Remove Parent Tags on Tag Deletion


clloyd:
When you delete a tag, if that tag is referred to as a parent by any other tags, the reference will be left "dangling"

PR #207, opened 2016-03-29
Add Support route to remove Dangling Parents

clloyd:
We seem to have inherited some dangling parents (tags that are listed as parents but don't actually exist) from R2

This is a support route to remove them. There's further piece of work to check if future parents need to be removed on Tag Deletion - #206


currysoup:
:+1:

PR #208, opened 2016-03-29
Blocking contributors being batch tagged

currysoup:
Batch tagging contributors 'works' but it doesn't play nicely with Composer so this will block it in the client side as a measure to prevent confusion.

If there's ever a desire for batch tagging to work with contributors then composer will need to be updated, for now just blocking it.


clloyd:
:+1:

PR #209, opened 2016-03-30
More detailed job system

currysoup:
Updated the job system to give more information to the user and reduce the amount of damage that can be done in the event of a step failure (or partial failure). Increased observability means I already found and fixed quite a few bugs while developing this PR. 

Most the diffs are just moving the steps into a different structure. Also removed the SQS system since that seemed to result in orphaned jobs. This makes taking a job a little more sophisticated since the node needs to be sure that no other node has picked up the job. I'm happy to add SQS back in if there's some way of making it more robust.


public/components/JobTable/JobTable.react.js
```diff
@@ -20,77 +21,220 @@ export default class JobTable extends React.Component {
...
+      return jobStatus;
+    }
 
-      if (step.type === 'TagRemovedCheck') {
-        return (
-          <span>
-            Confirming <b>{step.apiTagId}</b> has been removed from CAPI
-          </span>
-        );
+    prettyStepType(stepType) {
```
clloyd: All these should probably be moved out into a constants file

clloyd:
:+1:  Very small comment but looks awesome!

PR #211, opened 2016-03-31
Migrate Jobs to canonical Dynamo Tables

currysoup:


clloyd:
Would it be worthwhile setting this to a new jobs table? then we could switch over without downtime?

currysoup:
yeah i guess ;D

PR #213, opened 2016-03-31
Doesn't make sense to block people from filtering on contributors

currysoup:
Will continue to prevent people from batch adding/removing contributors for now.


clloyd:
:+1:

PR #214, opened 2016-04-01
Stop old content being included in batch tag operations

currysoup:
Grey out and prevent users from selecting content which isn't managed by composer in the batch tagger.

This is required since flex-feeds seemingly can't modify content which isn't managed by composer.

This can be removed in a few weeks when CAPI purges unmanaged content.

composerbatch


clloyd:
:+1:

PR #215, opened 2016-04-04
Sort on a field that actually exists

currysoup:
Even in the old job system the field was labelled 'started' :smile:


clloyd:
:+1:

PR #221, opened 2016-04-26
Add Migration endpoints for slugs, rename paths endpoint

clloyd:
As we are backfilling slugs and paths, this adds an endpoint for slugs and renames the paths endpoint


currysoup:
:+1:

PR #222, opened 2016-04-26
Add Section Expiry

clloyd:
This adds a very similar button to unexpiry that triggers an expiry.

Also as part of this PR I've removed the heading style which isn't needed from alot of places.


currysoup:
:+1: 🚀

PR #223, opened 2016-04-26
Used the frontend profile for logo upload locally.

steppenwells:
This mimics the sts assume role stuff as used in AWS


clloyd:
:+1:

PR #224, opened 2016-04-26
Sponsor logo styling fix

Calanthe:
A small styling fix to correctly show the images' dimensions.

Before:
screen shot 2016-04-26 at 15 18 14

After:
screen shot 2016-04-26 at 15 18 20

### DIFF
--- a/public/components/SponsorshipEdit/SponsorLogo.react.js
+++ b/public/components/SponsorshipEdit/SponsorLogo.react.js
@@ -60,7 +60,7 @@ export default class SponsorEdit extends React.Component {
     const imageAsset = this.props.logo.assets[0];
 
     return (
-      <div className="tag-edit__field">
+      <div className="tag-edit__field tag-edit__field-logo">
         <img src={imageAsset.imageUrl} className="tag-edit__field__image"/>
         <div className="tag-edit__image__info">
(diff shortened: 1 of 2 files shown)


steppenwells:
Looks good to me, but unfortunately I agreed with the tools team that we'd leave the plus 1'ing up to them so that they can make sure they're keeping up to date with what we're up to.

I'm sure @clloyd will be on this shortly.

public/components/SponsorshipEdit/SponsorLogo.react.js
```diff
@@ -60,7 +60,7 @@ export default class SponsorEdit extends React.Component {
     const imageAsset = this.props.logo.assets[0];
 
     return (
-      <div className="tag-edit__field">
+      <div className="tag-edit__field tag-edit__field-logo">
```
clloyd: We're using a BEM naming convention (ideally) in this project, so generally there should be a single class per thing... so as you need a new variant of the `tag-edit__field` class then you can create `tag-edit__field--logo` and then in the SASS do something similar to `%tag-edit__image__add`

clloyd:
Just got one small naming convention comment

Calanthe:
You are right @clloyd. I just committed those changes.

clloyd:
👍

PR #226, opened 2016-05-05
Adds podcast category support

clloyd:
Adds podcast categories,

Note: these are represented in the thrift definition as a list, but only a single main and sub category is applicable in TagManager. Hence the List creation in asThrift method

Tested on CODE and confirmed appears correctly in CAPI


currysoup:
So long as we're fine with the client defining the category/subcategory strings then this is :+1:

PR #228, opened 2016-05-09
Just a few little improvements...

lps88:
Fix spacings in top of sponsorships table, and navigation to the sponsorship page (so the back button works as expected)


clloyd:
:+1:

PR #229, opened 2016-05-13
Tag Manager FFF (Flaw Friday Fixes)

clloyd:
Paths are trimmed when calculated (no more tag-name--- for paths)
Sections are mandatory for series tags


hmgibson23:
👍

PR #230, opened 2016-05-13
Sponsorship list page now requires the commercial tags permission

steppenwells:
Looks like I missed a route when adding the new permission. (🐑ish)


clloyd:
👍

clloyd:
@steppenwells Still needed?

PR #231, opened 2016-05-13
Enable audit streaming to Tools Audit Server

clloyd:
This adds support for @piuccio 's auditing setup. When enabled (All env's apart from DEV) when modifying a Tag or Section, a kinesis message is sent to the stream in the Fronts-CMS account.

cc @steppenwells

### DIFF
--- a/app/model/SectionAudit.scala
+++ b/app/model/SectionAudit.scala
@@ -1,11 +1,12 @@
 package model
 
 import com.amazonaws.services.dynamodbv2.document.Item
+import com.gu.auditing.model.v1.{App, Notification}
 import com.gu.pandomainauth.model.User
 import org.joda.time.DateTime
 import play.api.Logger
 import play.api.libs.functional.syntax._
-import play.api.libs.json.{Json, JsPath, Format}
+import play.api.libs.json.{Format, JsPath, Json}
 import repositories.SectionRepository
 
 import scala.util.control.NonFatal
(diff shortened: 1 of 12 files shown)


app/model/SectionAudit.scala
```diff
@@ -20,6 +21,16 @@ case class SectionAudit(
                      sectionSummary: SectionSummary
                    ) {
   def toItem = Item.fromJSON(Json.toJson(this).toString())
+
+  def asAuditingThrift = Notification(
+    app = App.TagManager,
+    operation = operation,
+    userEmail = s"${user.replace(" ", ".").toLowerCase}@guardian.co.uk",
```
piuccio: is `user.email` not available? Some people have funny email, it's best to use whatever you use to authenticate against panda

app/model/SectionAudit.scala
```diff
@@ -20,6 +21,16 @@ case class SectionAudit(
...
+    date = date.toString,
+    resourceId = Some(sectionId.toString),
+    message = Some(s"${description}. Section: ${sectionSummary.toString}"),
+    shortMessage = Some(s"${description} Section ID: ${sectionId.toString}")
```
piuccio: it's up to you to decide what to do with `message` and `shortMessage`, but maybe you only need `Some(s"${description}")`, the `sectionId` is already inside `resourceId`

piuccio:
looks good

app/model/SectionAudit.scala
clloyd: Unfortunately not at the moment (I've added a flaw friday item for me to switch to using that rather than name to identify users with our job system)

clloyd: Going to make this change as a separate PR now

PR #234, opened 2016-06-06
Add Access-Control-Allow-Origin for composer-secondary.

jennysivapalan:
Note composer-secondary exists for CODE but not yet PROD. I'm happy to wait till PROD exists before merging but I don't think it will be a problem to merge this before.


clloyd:
👍

PR #235, opened 2016-06-20
Allow batch and merge tagging of contributors

clloyd:
This needs to be deployed after https://github.com/guardian/flexible-content/pull/2607

PR #237, opened 2016-06-20
Missed Contributor check on MergeTagCommand

clloyd:
This removes the contributor check on the MergeTagCommand (see #235 )


akash1810:
👍

PR #238, opened 2016-07-05
Add support for "Hosted Content" sub-type of paid content tag

clloyd:
This adds the new subtype, a new PaidContentInformationEdit component and allows the use of parameters on the image upload support endpoint to validate other sizes of images.


steppenwells:
👍

PR #239, opened 2016-07-29
Sw hosted content paths

steppenwells:
Ensures that all hosted content campaign tags are created wit an "advertiser-content/" path prefix.

Also ensures that microsites are always created for hosted content tags.

also makes path check for tracking tags work correctly.

### DIFF
--- a/public/components/TagEdit/formComponents/TagName.react.js
+++ b/public/components/TagEdit/formComponents/TagName.react.js
@@ -118,6 +118,16 @@ export default class TagNameEdit extends React.Component {
       return 'tracking/' + trackingTypeName + '/';
     }
 
+    // Paid content with sub type of hosted exception
+
+    if (this.props.tag.type === tagTypes.paidContent.name
+      && this.props.tag.paidContentInformation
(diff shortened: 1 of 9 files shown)


public/components/TagEdit/formComponents/TagName.react.js
```diff
@@ -118,6 +118,16 @@ export default class TagNameEdit extends React.Component {
       return 'tracking/' + trackingTypeName + '/';
     }
 
+    // Paid content with sub type of hosted exception
+
+    if (this.props.tag.type === tagTypes.paidContent.name
+      && this.props.tag.paidContentInformation
+      && this.props.tag.paidContentInformation.paidContentType
```
clloyd: This line technically isn't required as if `.paidContentInformation` exists then `.paidContentType` would be `undefined` if it doesn't exist

clloyd:
👍

public/components/TagEdit/formComponents/TagName.react.js
steppenwells: good point.

PR #240, opened 2016-07-29
Sw hypermedia subtype

steppenwells:
Adds subtype support to hypermedia api.

field turns up in results when populated

results can be filtered by subtype

supports both pain content and tracking tag subtypes.


clloyd:
👍

PR #242, opened 2016-08-05
Update react + dependencies, added colour picker

clloyd:
This updates react to v15 bringing a number of [benefits](https://facebook.github.io/react/blog/2016/04/07/react-v15.html)

As part of this bump I've:
 - refactored the routes to be compatible with newest react-router (moving to new routes.js)
 - removed the dependency on `history` and replaced with `browserHistory` found in `react-router`
 - Added colour picker component


steppenwells:
👍

PR #244, opened 2016-08-10
Fix for default state being undefined on input fields

clloyd:
This resulted in fields not being correctly blanked when reset tag action is performed


clloyd:
cc @steppenwells

steppenwells:
Looks amazing 👍

PR #245, opened 2016-08-16
Don't limit the size of images uploaded when tag type is Hosted Content.

clloyd:
@steppenwells

### DIFF
--- a/public/components/SponsorshipEdit/SponsorEdit.react.js
+++ b/public/components/SponsorshipEdit/SponsorEdit.react.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import SponsorLogo from './SponsorLogo.react';
+import {PAID_HOSTEDCONTENT_TYPE} from '../../constants/paidContentTagTypes';
 
 export default class SponsorEdit extends React.Component {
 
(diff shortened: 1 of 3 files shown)


public/components/SponsorshipEdit/SponsorEdit.react.js
```diff
@@ -37,6 +38,9 @@ export default class SponsorEdit extends React.Component {
       return false;
     }
 
+    const logoWidth = this.props.paidContentTagType === PAID_HOSTEDCONTENT_TYPE.value ? false : 140;
+    const logoHeight = this.props.paidContentTagType === PAID_HOSTEDCONTENT_TYPE.value ? false : 90;
```
steppenwells: I guess this works, but it looks kind of ghastly when you're used to stronger typing.

steppenwells:
👍

public/components/SponsorshipEdit/SponsorEdit.react.js
clloyd: It's either this or add more logic into the support method on the server, and pass tag information with the request... I know what you mean though

PR #251, opened 2016-09-14
Sw move paidcontent spons to section

steppenwells:
2 things here:

migration endpoint to flatten a sponsored tag to paid content.

updates the update tag command to trigger indexes of tags and sections correctly when the sponsorship updates. This fixes problems propagating changes to sponsorships seen in the wild.


clloyd:
👍

PR #252, opened 2016-09-14
remove guard preventing paid content migration happening when type is…

steppenwells:
… not paid content.

this is how we're fixing missed stuff so it's no longer relevant


steppenwells:
:+1: done interactively with @clloyd

PR #254, opened 2016-09-22
fire clash validation on all updates to sponsorship

steppenwells:
not just targeting changes.

This means that a sponsorship can be valid after changes to sponsor name etc. This change is not applied to create which will always need targeting configured.


steppenwells:
that was quick.

PR #268, opened 2016-12-05
include paid content tag's sponsorship in hypermedia api response

steppenwells:
even if it has expired.

This endpoint is used by campaign central


clloyd:
I'd keep an eye on dynamo after this update! :(

PR #270, opened 2016-12-07
Sw make paid content type mandatory

steppenwells:
2 bits: the frontend validations now ensure the subtype is populated, and a migration endpoint to fix existing problematic paid content tags.

### DIFF
--- a/app/controllers/Migration.scala
+++ b/app/controllers/Migration.scala
@@ -115,7 +115,7 @@ object Migration extends Controller with PanDomainAuthActions {
 
     val missingSubType = paidContentTags.filter{t =>
       t.paidContentInformation.isEmpty ||
-        t.paidContentInformation.get.paidContentType.trim == null ||
+        t.paidContentInformation.get.paidContentType == null ||
         t.paidContentInformation.get.paidContentType.trim == ""
     }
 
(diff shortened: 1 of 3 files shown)


app/controllers/Migration.scala
```diff
@@ -103,4 +106,35 @@ object Migration extends Controller with PanDomainAuthActions {
...
+  def addMissingPaidContentTypes() = (APIAuthAction andThen TriggerMigrationPermissionsCheck) {
+    implicit val username = Some("sponsorship Migration")
+
+    val paidContentTags = TagLookupCache.search(
+      TagSearchCriteria(types = Some(List("PaidContent")))
+    )
+
+    val missingSubType = paidContentTags.filter{t =>
+      t.paidContentInformation.isEmpty ||
+        t.paidContentInformation.get.paidContentType.trim == null ||
```
kelvin-chappell: Surely `trim` can never be null?  This would throw an NPE.

app/controllers/Migration.scala
```diff
@@ -103,4 +106,35 @@ object Migration extends Controller with PanDomainAuthActions {
...
+        t.paidContentInformation.get.paidContentType.trim == ""
+    }
```
kelvin-chappell:
```
... = paidContentTags.filterNot { t =>
  t.paidContentInformation.exists(_.paidContentType.nonEmpty)
}
```

app/controllers/Migration.scala
steppenwells: tis a good point.

PR #273, opened 2017-02-10
Improve thread pool handling

sihil:
Very large mapping manager changes have been causing problems recently due to long running tasks having been run in the main play threadpool. This pool defaults to having one worker per core - which is perfect for a 100% reactive application, but tag manager goes off and does a lot of DB work (and in the case of mapping updates executes a blocking task processing large CAPI result sets, pushing the results onto a kinesis stream).

This PR moves most potentially blocking code into a Future with a new more appropriate execution context - the main target was anything run within a `Command`. 
 - A new `Contexts` object has been created as a home to a new `tagOperationContext`. I've also moved the CAPI context into the same place (using Play's akka stuff deals with cleanup itself)
 - The `Command` trait's `process` method now returns a `Future[Option[T]]` instead of an `Option[T]`
 - Every command has been modified to wrap in a future that executes using the `tagOperationContext`
 - All calls within actions to command processes have now been modified to use `async`
 - All existing imports of the default scala execution context have been removed and replaced with the default play execution context or the new `tagExecutionContext` as appropriate

As much as possible I've not modified the existing code structure, simply lifted the existing flow into one that uses Futures and maps over Futures. There were a couple of places where this was trickier (e.g. multiple process results being dealt with in one location).

I've also bumped the `scalaVersion` and tidied up some compiler warnings.

The tests still pass, but that seems to be because there are not any tests 👼

### DIFF
--- a/app/controllers/App.scala
+++ b/app/controllers/App.scala
@@ -6,10 +6,10 @@ import play.api.mvc.{Action, Controller}
 import repositories.{TagSearchCriteria, TagRepository}
 import play.api.Logger
 import services.Config
-import com.gu.tagmanagement.{TagType}
+import com.gu.tagmanagement.TagType
 import permissions._
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import play.api.libs.concurrent.Execution.Implicits._
 import scala.collection.mutable.ListBuffer
 
 object App extends Controller with PanDomainAuthActions {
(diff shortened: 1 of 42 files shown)


currysoup:
:+1:

philmcmahon:
Tested on CODE with minor tag updates and batch tag operations. Changes seemed to be happily going through to CAPI.

PR #275, opened 2017-05-04
Validate external names cannot have a leading space

currysoup:
Should prevent the issue which recently cropped up where people cannot find a tag in composer because it's external name has a leading space.

I presume this is caused by copy-pasting from other documents or something...

### DIFF
--- a/public/util/validateTag.js
+++ b/public/util/validateTag.js
@@ -94,7 +94,7 @@ function validateTrackingTag(tag) {
 }
 
 function validateTagName(tag) {
-  let errors = [];
+  const errors = [];
   if (tag.externalName && tag.externalName.endsWith(' ')) {
     errors.push({
       fieldName: 'externalName',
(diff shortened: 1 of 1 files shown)


public/util/validateTag.js
```diff
@@ -94,14 +94,22 @@ function validateTrackingTag(tag) {
 }
 
 function validateTagName(tag) {
+  let errors = [];
```
akash1810: can be a `const`

currysoup:
I remember having a conversation about if we should just `trim` when I added trailing space prevention. We decided not to do that, but neither me nor David can remember why... It might be related to not making presumptions about what the users want and potentially changing the external name as a side effect of some other edit. If we can be sure the external name changing wouldn't be a problem then I think we can just add trimming.

There are currently 13 pieces of content which begin with a space, and a some unknown amount with a trailing space.

PR #280, opened 2017-05-17
dev-nginx provided do not exist

kant:
Gives me a 404 error


jharewinton:
https://github.com/guardian/dev-nginx does exist. It's a private Guardian repo

sihil:
Hi @kant
This is a developed in the open Guardian repo. Whilst we are happy for others to take an interest in what we do and deliberately choose to work in the open, this is not open source software and we don't really intend to support it for an audience outside our developer team. As @jharewinton says this repo does exist but is private as it contains data not suitable for being open.
You can read more about our approach for developing in the open at https://www.theguardian.com/info/developer-blog/2014/nov/28/developing-in-the-open.
Thanks for taking an interest!

PR #285, opened 2017-08-22
Tag search pagination

Reettaphant:
There was some code that in the controller that was reading the page size from the query from the client which didn't seem right so I added it to the config instead. If we wanted to enforce consistency in the display of the pagination panels, `PAGE_NAV_SPAN`  could go there as well and it could be used by the batch tagging component.

### DIFF
--- a/app/services/Config.scala
+++ b/app/services/Config.scala
@@ -112,6 +112,8 @@ sealed trait Config {
   def frontendBucketWriteRole: Option[String] = None
   def auditingKinesisWriteRole: Option[String] = None
   def enableAuditStreaming: Boolean = true
+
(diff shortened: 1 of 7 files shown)


app/services/Config.scala
```diff
@@ -112,6 +112,8 @@ sealed trait Config {
   def frontendBucketWriteRole: Option[String] = None
   def auditingKinesisWriteRole: Option[String] = None
   def enableAuditStreaming: Boolean = true
+
+  def tagSearchPageSize = 25
```
Reettaphant: I think it should be in some configuration file instead of just being passed in by the component and the configuration gets passed in by the server to that's why.

PR #287, opened 2017-09-02
remove auditing streams from tagmanager

currysoup:
No longer using audit streams so we can murderize them here


currysoup:
No longer required

rest in peace, little brancharoo

PR #288, opened 2017-09-04
Fixes

Reettaphant:
Fixes the two problems currently fixed in the temp branch here (https://github.com/guardian/tagmanager/pull/287)

- tag mappings are getting tags from the new api response correctly
- auditing cross account role updated to an existing one so that auditing events are going through again

### DIFF
--- a/app/services/Config.scala
+++ b/app/services/Config.scala
@@ -253,6 +253,6 @@ class ProdConfig extends Config {
     "https://campaign-central.gutools.co.uk")
 
   override def frontendBucketWriteRole: Option[String] = Some("arn:aws:iam::642631414762:role/composerWriteToStaticBucket")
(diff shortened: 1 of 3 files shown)


app/services/Config.scala
```diff
@@ -207,7 +207,7 @@ class CodeConfig extends Config {
     "https://campaign-central.local.dev-gutools.co.uk")
 
   override def frontendBucketWriteRole: Option[String] = Some("arn:aws:iam::642631414762:role/composerWriteToStaticBucket")
-  override def auditingKinesisWriteRole: Option[String] = Some("arn:aws:iam::163592447864:role/auditing-CrossAccountKinesisAccess-CC5UXEHZNP5M")
+  override def auditingKinesisWriteRole: Option[String] = Some("arn:aws:iam::163592447864:role/auditing-kinesis-cross-ac-CrossAccountKinesisAcces-PI8TASUEN2FE")
```
currysoup: Only the `CODE` role changed?

PR #292, opened 2017-09-27
Single CPU hack

sihil:
Tag manager suffers deadlocks on single CPU machines. This should fix that.

The real fix is to do a proper analysis of all blocking IO and ensure it's all on the correct execution context (or shifted to an async approach).


sihil:
Hmmm. Confusing.

PR #293, opened 2017-10-05
make images display even if the url is invalid

currysoup:
Previously we had to have a valid URL to display an image - this was problematic for images which had been migrated from R2.

### DIFF
--- a/public/components/TagEdit/formComponents/TagImageEdit.react.js
+++ b/public/components/TagEdit/formComponents/TagImageEdit.react.js
@@ -70,7 +70,6 @@ export default class TagImageEdit extends React.Component {
   }
 
   fetchMetadata(imageUrl) {
-
     //Validate before any requests happen
     if (!validateImageUrl(imageUrl)) {
       this.setState({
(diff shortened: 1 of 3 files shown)


public/components/TagEdit/formComponents/TagImageEdit.react.js
```diff
@@ -111,7 +110,7 @@ export default class TagImageEdit extends React.Component {
     }
 
     if (this.state.metadataFetchStatus === FETCH_STATES.success) {
-      return (<div className="tag-edit__image__add--success" onClick={this.addImage.bind(this)}><i className="i-tick-green" />Add Image</div>);
+      return (<div className="tag-edit__image__add--success" onClick={this.addImage.bind(this)}><i className="i-tick-green" />Click here to add the image</div>);
```
currysoup: Expanded the text here because some newer users to tag manager weren't aware that you had to click the `Add Image` text to add the image. Which is fair enough since it doesn't look like a button at all - it just looks like a notification that your image has been added.

currysoup:
Also @paperboyo this fixes the problem we saw the other day when working through the large bylines

PR #296, opened 2017-10-19
Allow podcasts to have an optional podcast type

annebyrne:
iOS 11 gives podcast level information about type. This impacts how episodes are displayed to users, but is an optional field as `episodic` will be the default.

### DIFF
--- a/app/model/PodcastMetadata.scala
+++ b/app/model/PodcastMetadata.scala
@@ -13,7 +13,8 @@ case class PodcastMetadata( linkUrl: String,
                             clean: Boolean = false,
                             explicit: Boolean = false,
                             image: Option[Image] = None,
-                            categories: Option[PodcastCategory] = None
+                            categories: Option[PodcastCategory] = None,
+                            podcastType: Option[String] = None
 ) {
 
   def asThrift = ThriftPodcastMetadata(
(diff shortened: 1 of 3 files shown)


app/model/PodcastMetadata.scala
```diff
@@ -38,6 +40,7 @@ case class PodcastMetadata( linkUrl: String,
     <explicit>{this.explicit}</explicit>
     <image>{this.image.map(x => x.asExportedXml).getOrElse("")}</image>
     <categories>{this.categories.map(_.asExportedXml).getOrElse("")}</categories>
+    <type>{this.podcastType}</type>
```
regiskuckaertz: `this.podcastType.getOrElse("")`?

public/components/TagEdit/formComponents/series/PodcastMetadata.react.js
```diff
@@ -79,6 +96,14 @@ export default class PodcastMetadata extends React.Component {
           tag={this.props.tag}
           updateTag={this.props.updateTag}/>
         <div className="tag-edit__field">
+          <label className="tag-edit__label">Podcast Type</label>
+          <select value={newType || ""} onChange={this.updatePodcastType.bind(this)}>
```
shaundillon: The `.bind(this)` isn't necessary if you use arrow functions.

annebyrne:
Tested in `CODE` so merging :+1: 

@shaundillon the problem running it locally seemed to be running `setup.sh` before `start.sh`. not sure why that is 🤔

PR #297, opened 2017-11-02
Pillars

tomrf1:
Introduces a `Pillar Editor` tab at the top.
The user can set the pillar name and its sections.
The path has the form `pillar/<hyphenated-name>`

Updates are sent to kinesis for capi-consumption.

### TODO later:

- [ ] Reindexing
- [x] Deletes

### Pillars list:
picture 10

### Pillar editor:
picture 13

### DIFF
--- a/app/model/command/CreatePillarCommand.scala
+++ b/app/model/command/CreatePillarCommand.scala
@@ -1,0 +1,64 @@
+package model.command
+
+import com.gu.tagmanagement.{PillarEvent, PillarEventType}
+import model._
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Format, JsPath}
+import repositories._
+import CommandError._
+import play.api.Logger
+import services.KinesisStreams
+
+import scala.concurrent.Future
+import services.Contexts.tagOperationContext
+
+case class CreatePillarCommand(
+  name: String,
+  path: String,
+  sectionIds: Seq[String] = Nil) extends Command {
+
+  type T = Pillar
+
+  def process()(implicit username: Option[String] = None): Future[Option[Pillar]] = {
+
+    val pageIdFuture: Future[Long] = Future { try { PathManager.registerPathAndGetPageId(path) } catch {
+      case p: PathRegistrationFailed => PathInUse
+      case e: Throwable =>
+        Logger.warn(s"Error using pathmanager for $path: ${e.getMessage}", e)
+        throw e
+    }}
+
+    val pillarIdFuture = Future{Sequences.pillarId.getNextId}
+
+    for {
+      pageId <- pageIdFuture
+      pillarId <- pillarIdFuture
+    } yield {
+      val pillar = Pillar(
+        id = pillarId,
+        name = name,
+        path = path,
+        pageId = pageId,
+        sectionIds = sectionIds
+      )
+
+      val result = PillarRepository.updatePillar(pillar)
+
+      KinesisStreams.pillarUpdateStream.publishUpdate(pillar.id.toString, PillarEvent(PillarEventType.Update, pillar.id, Some(Pillar.asThrift(pillar))))
+
+      PillarAuditRepository.upsertPillarAudit(PillarAudit.created(pillar))
+
+      result
+    }
+  }
+}
+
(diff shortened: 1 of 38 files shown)


app/model/command/CreatePillarCommand.scala
```diff
@@ -0,0 +1,64 @@
...
+      PillarAuditRepository.upsertPillarAudit(PillarAudit.created(pillar))
+
+      result
+    }
+  }
+}
+
+object CreatePillarCommand {
+
+  implicit val createPillarCommandFormat: Format[CreatePillarCommand] = (
```
currysoup: Are we serializing the command?

app/repositories/PillarRepository.scala
```diff
@@ -0,0 +1,29 @@
...
+  def getPillar(id: Long): Option[Pillar] = {
+    Option(Dynamo.pillarTable.getItem("id", id)).map(Pillar.fromItem)
+  }
+
+  def updatePillar(pillar: Pillar): Option[Pillar] = {
+    try {
+      Dynamo.pillarTable.putItem(Pillar.toItem(pillar))
+      Some(pillar)
+    } catch {
+      case e: Error =>
```
currysoup: Should probably be NonFatal or something? `Error` in java is usually for things like out of heap space which we can't really handle.

app/repositories/PillarRepository.scala
```diff
@@ -0,0 +1,29 @@
...
+        Logger.warn(s"Error updating pillar ${pillar.id}: ${e.getMessage}", e)
+        None
+    }
+  }
+
+  def loadAllPillars: Iterable[Pillar] = Dynamo.pillarTable.scan().map(Pillar.fromItem)
+
+  def count: Int = Dynamo.pillarTable.scan().count(_ => true)
```
currysoup: This doesn't seem to be used anywhere?

app/repositories/PillarRepository.scala
tomrf1: true, though `Error` is currently used all over this app :)

app/model/command/CreatePillarCommand.scala
tomrf1: We deserialize in the controller, I suppose it doesn't need to be a `Format` but it's consistent with the rest of this app...

currysoup: Ah, apologies I just missed that :)

app/repositories/PillarRepository.scala
currysoup: Yeah, looks like some mistakes were made 😞 

I'll add it to our health trello

app/model/command/DeletePillarCommand.scala
```diff
@@ -0,0 +1,19 @@
...
+import services.{Contexts, KinesisStreams}
+
+import scala.concurrent.Future
+
+
+case class DeletePillarCommand(id: Long) extends Command {
+  override type T = Long
+
+  override def process()(implicit username: Option[String] = None): Future[Option[Long]] = Future {
+    KinesisStreams.pillarUpdateStream.publishUpdate(id.toString, PillarEvent(PillarEventType.Delete, id, None))
```
currysoup: Do we want to free up the path in path manager here?

Will this result in content being modified in CAPI?

tomrf1: pillars are not stored on the content. As requests come in capi adds the pillar to the content response based on the section.

public/components/Pillar/PillarEdit.react.js
```diff
@@ -0,0 +1,36 @@
...
+    render () {
+
+      if (!this.props.pillar) {
+        return false;
+      }
+
+      return (
+          <div className="tag-edit__input-group">
+              <div className="tag-edit__name">
+                  <label className="tag-edit__input-group__header">Name</label>
```
currysoup: Would be nice to render the `slugify`ed path here too so they can see how it will look prior to publishing and also be consistent with the tags and sections.

tomrf1:
Note - I've commented out the 'Pillar Editor' tab for now

PR #299, opened 2017-11-08
Add the ability to filter tags by the existence of a field

currysoup:
Currently only support contributorInfromation.bylineImage and ".largeBylineImage since this is all that's needed for now.

### DIFF
--- a/app/controllers/TagManagementApi.scala
+++ b/app/controllers/TagManagementApi.scala
@@ -51,12 +51,13 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
   }
 
   def searchTags = APIAuthAction { req =>
-    
+
     val criteria = TagSearchCriteria(
       q = req.getQueryString("q"),
       searchField = req.getQueryString("searchField"),
       types = req.getQueryString("types").map(_.split(",").toList),
-      referenceType = req.getQueryString("referenceType")
+      referenceType = req.getQueryString("referenceType"),
+      hasFields = req.getQueryString("hasFields").map(_.split(",").toList)
     )
 
     val orderBy = req.getQueryString("orderBy").getOrElse("internalName")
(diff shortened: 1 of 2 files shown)


app/repositories/TagRepository.scala
```diff
@@ -108,6 +110,17 @@ case class TagSearchCriteria(
 
   private def typeFilter(types: List[String])(tags: List[Tag]) = tags.filter { t => types.contains(t.`type`.toLowerCase) }
 
+  private def hasFieldsFilter(fields: List[String])(tags: List[Tag]) = tags.filter { t =>
+    fields.forall {
+        case "contributorinformation.bylineimage" => t.contributorInformation.flatMap(_.bylineImage).isDefined
+        case "contributorinformation.largebylineimage" => t.contributorInformation.flatMap(_.largeBylineImage).isDefined
+        case unsupported => {
+          Logger.warn(s"Attempted to check if tag has field '$unsupported' which is not supported by the hasFieldsFilter")
```
mbarton: Could we communicate this failure back to the client?

app/controllers/TagManagementApi.scala
```diff
@@ -51,12 +51,13 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
   def searchTags = APIAuthAction { req =>
-    
+
     val criteria = TagSearchCriteria(
       q = req.getQueryString("q"),
       searchField = req.getQueryString("searchField"),
       types = req.getQueryString("types").map(_.split(",").toList),
-      referenceType = req.getQueryString("referenceType")
+      referenceType = req.getQueryString("referenceType"),
+      hasFields = req.getQueryString("hasFields").map(_.split(",").toList)
```
mbarton: Potentially needs a `trim`?

currysoup: Sure, I'll add that to `hasFields` and `types`!

app/repositories/TagRepository.scala
currysoup: Was previously just defaulted to true if it was unsupported... might just go back to that.

PR #301, opened 2017-12-21
Add contributor firstname and lastname fields to contributor information

7091lapS:
- Add new fields
- Add client side validation for lastName field
- Edit circle yaml to fix build

screen shot 2017-12-21 at 15 34 17

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ machine:
   java:
     version: oraclejdk8
   node:
-    version: 5.2.0
+    version: 6.9.0
   environment:
     NODE_ENV: production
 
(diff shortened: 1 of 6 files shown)


circle.yml
```diff
@@ -3,7 +3,7 @@ machine:
   java:
     version: oraclejdk8
   node:
-    version: 5.2.0
+    version: 6.9.0
```
mbarton: Was there a particular need to bump the version?

7091lapS: Yes, the latest version of master from ~25 days ago  had stopped building on circe ci (and was never deployed). Changing the node version fixes this.

PR #304, opened 2018-01-19
Capi preview IAM

tomrf1:
We are in the process of changing how we authorise access to CAPI preview. In the past we have used basic auth (in Concierge), with the usernames/passwords maintained by the CAPI team.
Instead we will use api-gateway to move authorisation out of concierge and avoid having to maintain usernames/passwords.

The api-gateway uses IAM authorisation. This means the app assumes a cross-account role, and signs each (preview) request using this role.

Note - this means you will now need capi credentials to run the app locally

tested in CODE

### DIFF
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -4,8 +4,11 @@ import java.io.File
 import java.util.UUID
 import java.util.concurrent.TimeUnit
 
+import com.amazonaws.auth.{AWSCredentialsProvider, AWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.services.s3.model._
-import com.squareup.okhttp.{Credentials, OkHttpClient, Request}
+import com.gu.contentapi.client.IAMSigner
+import com.squareup.okhttp.{Headers, OkHttpClient, Request}
 import model.command.CommandError._
 import model.command.{ExpireSectionContentCommand, UnexpireSectionContentCommand, UpdateTagCommand}
 import model.{DenormalisedTag, Image, ImageAsset}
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -17,8 +17,8 @@ scalacOptions ++= Seq(
 resolvers += "Guardian Bintray" at "https://dl.bintray.com/guardian/editorial-tools"
 
 lazy val dependencies = Seq(
-  "com.amazonaws" % "aws-java-sdk" % "1.10.33",
-  "com.amazonaws" % "amazon-kinesis-client" % "1.6.1",
+  "com.amazonaws" % "aws-java-sdk" % "1.11.259",
```
tomrf1: the new dependency (content-api-client-aws) has a more recent aws sdk version

app/controllers/Support.scala
```diff
@@ -20,10 +23,23 @@ import services.{AWS, Config, ImageMetadataService}
 
 import scala.concurrent.Future
 import scala.util.control.NonFatal
+import collection.JavaConverters._
 
 
 object Support extends Controller with PanDomainAuthActions {
 
+  private val capiPreviewCredentials: AWSCredentialsProvider = {
```
currysoup: Would it be possible to take this out of the repository rather than have a copy-paste?

app/services/Config.scala
```diff
@@ -27,9 +27,9 @@ sealed trait Config {
 
   def capiUrl: String = getRequiredStringProperty("capi.url")
   def capiKey: String = getRequiredStringProperty("capi.key")
-  def capiPreviewUrl: String = getRequiredStringProperty("capi.preview.url")
-  def capiPreviewUser: String = getRequiredStringProperty("capi.preview.username")
-  def capiPreviewPassword: String = getRequiredStringProperty("capi.preview.password")
+  def capiPreviewIAMUrl: String = getRequiredStringProperty("capi.preview.iamUrl")
+  def capiPreviewRole: String = getRequiredStringProperty("capi.preview.role")
```
currysoup: Do both these new variables have to be in the "hidden" S3 config or can they be in the regular in-code config?

Particularly the role name

tomrf1: the role arn is effectively replacing the username/password - I've been handling it as 'secret' config so far

PR #305, opened 2018-01-25
allow clicking batch tag links

currysoup:


public/components/ContentList/ContentListItem.js
```diff
@@ -71,7 +76,15 @@ export default class ContentListItem extends React.Component {
             <td>
               {this.props.content.fields && this.props.content.fields.isLive === "false" ? "Not Published" : moment(this.props.content.webPublicationDate).format('DD/MM/YYYY')}
             </td>
-            <td>{this.props.content.id}</td>
+            <td>
+              {
+                this.props.content.fields && this.props.content.fields.isLive === "true"
+                ?
+                  <span className="contentlist__link" onClick={this.openLink.bind(this)} target="_blank">{this.props.content.id}</span>
```
shaundillon: Can't this just be an actual link?

currysoup: Can't stop the propagation of the click event from selecting the row if we do that (as far as I know)

shaundillon: Ah, I get you. You can, by having an `onClick` and `href` and have the `onClick` function stop the propogation. Not great

PR #306, opened 2018-01-30
Fix tag deletions

tomrf1:
1. Stop using dispatch for capi preview requests as it's not working properly (by default the `content-api-client` uses okhttp, which we already use elsewhere in tagmanager).
2. Stop sending requests to capi with no content ids, e.g. `...&ids=` in `countOccurencesOfTagInContents`. This fails because the signing logic drops the invalid param, so the actual url doesn't match.
3. Also, only consume the response body stream once when logging error responses (it breaks otherwise)

### DIFF
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -1,17 +1,15 @@
 package controllers
 
 import java.io.File
+import java.net.URI
 import java.util.UUID
 import java.util.concurrent.TimeUnit
 
-import com.amazonaws.auth.{AWSCredentialsProvider, AWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
-import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.services.s3.model._
-import com.gu.contentapi.client.IAMSigner
-import com.squareup.okhttp.{Headers, OkHttpClient, Request}
 import model.command.CommandError._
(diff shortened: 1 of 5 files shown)


app/repositories/PathManager.scala
```diff
@@ -12,13 +12,13 @@ case class PathRemoveFailed(m: String) extends RuntimeException(m)
 
 object PathManager {
 
-  val httpClient = new OkHttpClient()
+  val httpClient = new OkHttpClient.Builder().connectTimeout(5, TimeUnit.SECONDS).build
```
mchv: why do you need to override the default timeout  will be useful in a comment

app/controllers/Support.scala
```diff
@@ -28,6 +26,8 @@ import collection.JavaConverters._
 
 object Support extends Controller with PanDomainAuthActions {
 
+  private val httpClient = new OkHttpClient.Builder().connectTimeout(5, TimeUnit.SECONDS).build
```
mchv: 👍  creating once make much more sense (reduce GC pressure, improve perfs)

mchv:
👍

app/repositories/PathManager.scala
tomrf1: I was just keeping the existing timeout used by this class

mchv: oh I see

PR #311, opened 2018-02-27
add perf logging

currysoup:
Experimental feature which we can hopefully adopt more broadly


franziskas:
@currysoup anything to salvage here?

PR #313, opened 2018-03-01
Basic check to prevent people uploaded jpegs as large bylines

currysoup:
Very simple extra check to prevent people from uploading JPEGs as a large byline image. From time to time whoever is making the tag will upload the same jpeg for both the small and large byline image.

It's important this doesn't happen now because the large bylines can end up in the paper.


paperboyo:
Tw qs:
- why did I miss it?
- why did you close it?

PR #314, opened 2018-03-08
Add ssm role to cloudformation

currysoup:
We already use an AMI with the ssm-agent:

https://github.com/guardian/tagmanager/blob/master/riff-raff.yaml#L18


franziskas:
@currysoup anything to salvage here?

currysoup:
Ah! Akash made a separate PR for this with the guardian policy :)

Closing

PR #317, opened 2018-04-27
Add new Campaign Tag with subtype of callouts

ajwl:
This creates a new Tag Type called "Campaign" with the subtype "Callout"
It will work in a similar way to the Tracking tag 

As part of the changes to introduce callouts in the frontend, we need a new way to attach them to individual articles. In Targeting, we want to create a new unique tag for each callout. It will be of type "campaign".

**Tag manager UI with Campaign options**

screen shot 2018-04-30 at 13 48 16

**How frontend will render the campaign (off the config object**
screen shot 2018-04-13 at 13 32 35

### DIFF
--- a/app/model/command/CreateTagCommand.scala
+++ b/app/model/command/CreateTagCommand.scala
@@ -72,6 +72,7 @@ case class CreateTagCommand(
                       isMicrosite: Boolean,
                       capiSectionId: Option[String] = None,
                       trackingInformation: Option[TrackingInformation] = None,
+                      campaignInformation: Option[CampaignInformation] = None,
                       preCalculatedPath: Option[String] = None, //This is used so path isn't calculated
                       sponsorship: Option[InlinePaidContentSponsorshipCommand] = None,
                       paidContentInformation: Option[PaidContentInformation] = None,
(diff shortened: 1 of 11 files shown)


ajwl:
@currysoup @tomrf1 - how does this look to you?

app/model/command/CreateTagCommand.scala
```diff
@@ -120,9 +121,14 @@ case class CreateTagCommand(
 
     val createdSectionId = if(createMicrosite) { sectionId } else { None }
 
+    val tagSubType: Option[String] =  `type` match {
+      case "Tracking" => trackingInformation.map(_.trackingType)
+      case "Campaign" => campaignInformation.map(_.campaignType)
+    }
+
```
ajwl: Not sure if this is the best way to do this. What would be more robust?

app/model/command/CreateTagCommand.scala
```diff
@@ -120,9 +121,14 @@ case class CreateTagCommand(
 
     val createdSectionId = if(createMicrosite) { sectionId } else { None }
 
+    val tagSubType: Option[String] =  `type` match {
```
tomrf1: do we need to match on `type` here? Or can we do:
`trackingInformation.map(_.trackingType) orElse campaignInformation.map(_.campaignType)`

ajwl:
Could probably reduce some replication by having a `subtype` react component, instead of one for tracking and one for campaigns. Let me know what you think

public/components/TagEdit/formComponents/campaigns/CampaignInformation.react.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    render () {
+
+        const selectCampaignType = this.props.tag.campaignInformation ? this.props.tag.campaignInformation.campaignType : undefined;
+
+        return (
+            <div className="tag-edit__input-group">
+            <label className="tag-edit__input-group__header">Campaign Information</label>
+        <div className="tag-edit__field">
```
tomrf1: the indentation has gone a bit wrong here :(

app/model/command/CreateTagCommand.scala
currysoup: I think what's there communicates intent more clearly

tomrf1:
@ajwl I should have mentioned this before - please make the model updates in CAPI (elasticsearch + porter) before deploying this change. Let me know if you need a hand

PR #318, opened 2018-05-01
FIX: for bug that broke tag saving

ajwl:
A switch statement without a default state broke tag saving. For tags without a subtype. 
Caused by adding a new tag in PR yesterday. 

This appears to solve the problem.


ajwl:
phew - this fix seems to work @currysoup

PR #320, opened 2018-05-03
CALLOUTS: Whitelist Targeting to allow to the app to create tags

ajwl:
We want Targeting to be able to:
1.  Create Tags in TagManager
2. Search for Tags in TagManger 
3. Delete tags  in TagManager (of the type campaign)

### Extract CORSable 
To do that, it needs to make POST request and to do that, it needs to overcome Cross Origin Restrictions. 
This PR extracts the `CORSable` function used elsewhere into its own helper.
Have created a short list of: `corsablePostDomains` - which are apps whitelisted to do this. Only Targeting is on the list for now. 

### Wrap 3 functions with CORSable
We now added the CORSable check to the `createTag`, `searchTag` and `deleteTag` function, which allows Targeting to now access this function to create tags. 

### Allow hmac users to pass the email permissions test 
The createTag request comes from the front end of Targeting.
The deleteTag and searchTag requests come from the backend of Targeting so use the HMAC convention. 
Given that HMAC already provides a layer of security, we allow authenticated hmac requests to use a default email address. A change in `panda-hmac` allows this to happen. Hence the version bump

### DIFF
--- a/app/controllers/TagManagementApi.scala
+++ b/app/controllers/TagManagementApi.scala
@@ -11,6 +11,8 @@ import play.api.mvc.{Controller, Result}
 import repositories._
 import play.api.libs.concurrent.Execution.Implicits._
 import services.Config
+import helpers.CORSable
+import services.Config.conf
 
 import scala.concurrent.Future
 
@@ -37,21 +39,22 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
     }
   }
 
-  def createTag() = (APIAuthAction andThen CreateTagPermissionsCheck).async { req =>
-    implicit val username = Option(req.user.email)
-    req.body.asJson.map { json =>
-      json.as[CreateTagCommand].process.map { result =>
-        result.map{t => Ok(Json.toJson(DenormalisedTag(t))) } getOrElse NotFound
-      } recover {
-        commandErrorAsResult
+  def createTag = CORSable(conf.corsablePostDomains: _*) {
+    (APIAuthAction andThen CreateTagPermissionsCheck).async { req =>
+      implicit val username = Option(req.user.email)
+      req.body.asJson.map { json =>
+        json.as[CreateTagCommand].process.map { result =>
+          result.map { t => Ok(Json.toJson(DenormalisedTag(t))) } getOrElse NotFound
+        } recover {
+          commandErrorAsResult
+        }
+      }.getOrElse {
+        Future.successful(BadRequest("Expecting Json data"))
       }
-    }.getOrElse {
-      Future.successful(BadRequest("Expecting Json data"))
     }
   }
 
   def searchTags = APIHMACAuthAction { req =>
-
     val criteria = TagSearchCriteria(
       q = req.getQueryString("q"),
       searchField = req.getQueryString("searchField"),
@@ -65,16 +68,16 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
     val tags = TagLookupCache.search(criteria)
 
     val orderedTags: List[Tag] = orderBy match {
-      case("internalName") => tags.sortBy(_.internalName)
-      case("externalName") => tags.sortBy(_.externalName)
-      case("path") => tags.sortBy(_.path)
-      case("id") => tags.sortBy(_.id)
-      case("type") => tags.sortBy(_.`type`)
-      case(_) => tags.sortBy(_.comparableValue)
+      case ("internalName") => tags.sortBy(_.internalName)
+      case ("externalName") => tags.sortBy(_.externalName)
+      case ("path") => tags.sortBy(_.path)
+      case ("id") => tags.sortBy(_.id)
+      case ("type") => tags.sortBy(_.`type`)
+      case (_) => tags.sortBy(_.comparableValue)
     }
 
     val page = req.getQueryString("page").getOrElse("1").toInt
-    val pageSize = req.getQueryString("pageSize").map( _.toInt).getOrElse(Config().tagSearchPageSize)
+    val pageSize = req.getQueryString("pageSize").map(_.toInt).getOrElse(Config().tagSearchPageSize)
 
     val startIndex = (page - 1) * pageSize
     val paginatedTagResults = orderedTags.drop(startIndex).take(pageSize)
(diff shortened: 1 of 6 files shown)


app/controllers/TagManagementApi.scala
```diff
@@ -227,12 +231,14 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
   }
 
-  def deleteTag(id: Long) = (APIAuthAction andThen DeleteTagPermissionsCheck).async { req =>
-    implicit val username = Option(req.user.email)
-    (new DeleteTagCommand(id)).process.map{ result =>
-      result.map{t => NoContent } getOrElse NotFound
-    } recover {
-      commandErrorAsResult
+  def deleteTag(id: Long) =  CORSable(conf.corsablePostDomains: _*) {
+    (APIAuthAction andThen DeleteTagPermissionsCheck).async { req =>
```
currysoup: If we're planning to have a background process in Targeting delete tags in tag manager then we might need to make this an `APIHMACAuthAction`

ajwl:
@currysoup Sam do I need to wrap the deleteTag and searchTag functions in the Corsable function, given that they are accessed from the backend via hmac and not the frontend via cookies etc?

ajwl:
Also - when you have time to review, that would be great @currysoup

app/controllers/TagManagementApi.scala
```diff
@@ -37,50 +39,54 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
+        Future.successful(BadRequest("Expecting Json data"))
       }
-    }.getOrElse {
-      Future.successful(BadRequest("Expecting Json data"))
     }
   }
 
-  def searchTags = APIHMACAuthAction { req =>
+  def searchTags = CORSable(conf.corsablePostDomains: _*) {
+    APIHMACAuthAction { req =>
```
ajwl: Sorry for all the horrible spacing diffs here - the only change is this line: `def searchTags = CORSable(conf.corsablePostDomains: _*) {`

app/controllers/TagManagementApi.scala
```diff
@@ -227,12 +233,14 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
 
-  def deleteTag(id: Long) = (APIAuthAction andThen DeleteTagPermissionsCheck).async { req =>
-    implicit val username = Option(req.user.email)
-    (new DeleteTagCommand(id)).process.map{ result =>
-      result.map{t => NoContent } getOrElse NotFound
-    } recover {
-      commandErrorAsResult
+  def deleteTag(id: Long) =  CORSable(conf.corsablePostDomains: _*) {
+    (APIHMACAuthAction andThen DeleteTagPermissionsCheck).async { req =>
+      implicit val username = Option(req.user.email)
```
LATaylor-guardian: As above.

app/controllers/TagManagementApi.scala
```diff
@@ -227,12 +233,14 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
+      (new DeleteTagCommand(id)).process.map { result =>
```
LATaylor-guardian: Do you need the `new` keyword here for `DeleteTagCommand`? Is it a case class or a class?

app/permissions/PermissionActionCheck.scala
```diff
@@ -13,12 +13,16 @@ trait PermissionActionFilter extends ActionFilter[UserRequest] {
...
-        println("FAILED")
-        Some(Results.Unauthorized)}
+    if(request.user.email == "hmac-authed-service") {
+      Future.successful(None)
+    } else {
+      testAccess(request.user.email).map {
+        case PermissionGranted => None
+        case PermissionDenied =>
+          Logger.info(s"user not authorized to $restrictedAction")
+          println("FAILED")
```
LATaylor-guardian: Is this println still needed?

app/services/Config.scala
```diff
@@ -163,6 +164,9 @@ class DevConfig extends Config {
     "https://targeting.local.dev-gutools.co.uk",
     "https://campaign-central.local.dev-gutools.co.uk"
   )
+  override def corsablePostDomains: Seq[String] = Seq(
+    "https://targeting.local.dev-gutools.co.uk"
```
LATaylor-guardian: We can pull these URLs out into a `val` and reference them in the function so that they are not created every time the function is invoked.

app/controllers/TagManagementApi.scala
ajwl: it is indeed a case class, so doesn't need `new`

app/permissions/PermissionActionCheck.scala
ajwl: Seems like no, if there's a log, I'll take it out

app/controllers/TagManagementApi.scala
```diff
@@ -37,50 +39,54 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
...
-  def createTag() = (APIAuthAction andThen CreateTagPermissionsCheck).async { req =>
-    implicit val username = Option(req.user.email)
-    req.body.asJson.map { json =>
-      json.as[CreateTagCommand].process.map { result =>
-        result.map{t => Ok(Json.toJson(DenormalisedTag(t))) } getOrElse NotFound
-      } recover {
-        commandErrorAsResult
+  def createTag = CORSable(conf.corsablePostDomains: _*) {
+    (APIAuthAction andThen CreateTagPermissionsCheck).async { req =>
+      implicit val username = Option(req.user.email)
```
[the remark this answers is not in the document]
ajwl: How would that look Luke? Do I simply remove the word `implicit` ?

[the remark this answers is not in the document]
ajwl: Also I see the whole file is done in this way... should I change all of them?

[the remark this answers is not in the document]
ajwl: After talking to @currysoup about this, have decided that while good practice, this improvment would be outside the scope of the PR since all the other methods on the `TagManagementApi` work in this way. So would require a fair amount of clean up

ajwl:
@currysoup and in fact we didn't need it! removed.
You feeling happy with this PR..? Reckon we could meeeerge?

ajwl:
For future reference - the `new` keyword on [this line](https://github.com/guardian/tagmanager/blob/master/app/controllers/TagManagementApi.scala#L232): 
```
(new DeleteTagCommand(id)).process.map { result =>
```
Somehow allows contexts to function correctly. ✨ Removing it breaks functionality.

ajwl:
[Campaigns: change Targeting so that a new tag is generated for each new callout.](https://trello.com/c/2v0gK4IB/291-campaigns-change-targeting-so-that-a-new-tag-is-generated-for-each-new-callout)

PR #321, opened 2018-05-15
Basic check to prevent people uploaded jpegs as large bylines

currysoup:
Large byline images must have transparency so have to be pngs


currysoup:
Thanks buddy!

PR #331, opened 2018-08-14
Podcasts: Add a field for Google Podcasts & Spotify Links

ajwl:
We want to send over Google Podcasts links and Spotify links, in addition to the iTunes url.

This updates the interface for Podcast series and the backend too.


ajwl:
Branch deployed to CODE

tomrf1:
👍 looks good, worth checking it gets into CODE capi

ajwl:
@tomrf1 have added the changes
Is up now in CODE: https://tagmanager.code.dev-gutools.co.uk/tag/73818
Tho having trouble with local version of TM

tomrf1:
@ajwl thanks, I'll test this with CODE capi and then we can merge

PR #334, opened 2018-09-17
Typo on #44. Duplicate word on #16

kant:
Minor fix on string #18


franziskas:
@kant thanks for fixing this, good spot!

PR #339, opened 2018-10-07
Allow users to create a basket of changes in batch tagger

currysoup:


app/model/command/BatchTagCommand.scala
```diff
@@ -2,32 +2,38 @@ package model.command
...
+    val hasIntersections =
+      toTopList.intersect(toRemove).nonEmpty ||
+      toTopList.intersect(toAddToBottom).nonEmpty ||
+      toAddToBottom.intersect(toRemove).nonEmpty
+
+    if (hasIntersections) {
+      IntersectingBatchTags
     }
 
+    val tops = toAddToTop.map(TagRepository.getTag(_).getOrElse(TagNotFound))
```
LATaylor-guardian: Small thing but you can probably pull `TagRepository.getTag(_).getOrElse(TagNotFound)` out into a def and call from each of these vals.

app/model/command/BatchTagCommand.scala
```diff
@@ -2,32 +2,38 @@ package model.command
...
-import play.api.libs.functional.syntax._
-import play.api.libs.json.{Format, JsPath}
+import org.cvogt.play.json.Jsonx
 import repositories._
 import services.Contexts
 
 import scala.concurrent.Future
 
-case class BatchTagCommand(contentIds: List[String], tagId: Long, operation: String) extends Command {
+case class BatchTagCommand(contentIds: List[String], toAddToTop: Option[Long], toAddToBottom: List[Long], toRemove: List[Long]) extends Command {
```
LATaylor-guardian: Can `toAddToTop: Option[Long], toAddToBottom: List[Long], toRemove: List[Long]` be modelled as a single case class called `Operations`?

app/model/jobs/steps/ModifyContentTags.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import scala.language.postfixOps
+
+// The 'op' argument of this case class should really be the BatchTagOperation type but serializing enums to JSON isn't very easy with the
+// current version of play and enumeratim
+case class ModifyContentTags(tag: Tag, section: Option[Section], contentIds: List[String], op: String, `type`: String = ModifyContentTags.`type`, var stepStatus: String = StepStatus.ready, var stepMessage: String = "Waiting", var attempts: Int = 0) extends Step {
+  override def process = {
+    contentIds foreach { contentPath =>
+      val taggingOperation = TaggingOperation(
+        operation = OperationType.valueOf(op.replace("-", "")).get,
```
LATaylor-guardian: Is `.get` safe here?

app/model/jobs/steps/ModifyContentTags.scala
```diff
@@ -0,0 +1,70 @@
...
+        contentPath = contentPath,
+        tag = Some(TagWithSection(tag.asThrift, section.map(_.asThrift)))
+      )
+      KinesisStreams.taggingOperationsStream.publishUpdate(contentPath.take(128), taggingOperation)
```
LATaylor-guardian: Why do we take `128` here? Could pull out in a val for readability to explain this.

app/model/jobs/steps/ModifyContentTags.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  override def waitDuration: Option[Duration] = {
+    Some(5 seconds)
+  }
+
+  override def check: Boolean = {
+    val count = ContentAPI.countOccurencesOfTagInContents(contentIds, tag.path)
+
+    // As stated above, once we get the BatchTagOperation enum type serializable we can match against a type here instead of a string
```
LATaylor-guardian: 👍

app/model/jobs/steps/ModifyContentTags.scala
```diff
@@ -0,0 +1,70 @@
...
+    val expected = op match {
+      case "remove" => 0
+      case _ => contentIds.length
+    }
+
+    Logger.info(s"Checking batch tag operations. Expected=$expected Actual=$count")
+    if (count == expected) {
```
LATaylor-guardian: This can simply be written as `count == expected`.

scripts/start.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 npm run build-dev &
-AWS_PROFILE=composer sbt run
+AWS_PROFILE=composer sbt run -jvm-debug 5005
```
LATaylor-guardian: Did you mean to leave this here?

app/model/jobs/steps/ModifyContentTags.scala
currysoup: Should be, we pass the string via an enum that validates the input is one of `remove`, `add-to-top` or `add-to-bottom`. Sadly the thrift enums drop the dashes so we need the replace.

app/model/jobs/steps/ModifyContentTags.scala
currysoup: The online docs say that the maximum length of a kinesis partition key is 256 bytes and JVM strings are fixed 16 bytes wide so we limit the partition key length to 128 characters.

Weirdly enough the java doc contradicts this and states the maximum length is 256 "unicode characters" whatever that means.

app/model/jobs/steps/ModifyContentTags.scala
currysoup: We had a discussion about what was clearer here. I personally find Scala's lack of `return` statements and semicolons can make a tail-end boolean evaluation a tad confusing. Will change though.

scripts/start.sh
currysoup: Personally don't see the harm it does especially since `scripts/start.sh` is used just for development. It's convenient for the debugger to be available. :)

currysoup:
Just a notice - before this gets merged I'd like to do some comms with CP so they're prepared.

PR #340, opened 2018-10-10
replace curly quotes with straight ones in search queries

currysoup:
> The straights are at it again.

The ever reoccurring issue of fancy-pants variations on standard characters. Now we "ASCII fold" curly quotes when searching for tags

### DIFF
--- a/app/repositories/TagRepository.scala
+++ b/app/repositories/TagRepository.scala
@@ -4,9 +4,11 @@ import java.util.concurrent.atomic.AtomicReference
 
 import com.amazonaws.services.dynamodbv2.document.{Item, ScanFilter}
 import model.Tag
+import org.apache.commons.lang3.StringUtils
 import play.api.Logger
 import play.api.libs.json.JsValue
 import services.Dynamo
+
 import scala.collection.JavaConversions._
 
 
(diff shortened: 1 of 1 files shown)


paperboyo:
This is great, Mister!

In an ideal world:
- this would be an external config used by all Tools (and frontends!)
- should contain standard Unicode [normalisation routines](http://www.unicode.org/charts/normalization/) (e.g. [Canonical and Compatibility Equivalence](http://unicode.org/reports/tr15/#Canon_Compat_Equivalence))
- would contain config to automagically turn crude into proper while-you-type and disallow certain forms in certain fields (crappy quotes in Frontstool-edited headlines!)
- make me a sandwich. Preferably cheese Ploughman’s

For today:
– — − for -
“ for "
” for "
‘ for '
’ for '
… for ...

Also, see e.g. [here](https://github.com/guardian/flexible-content/blob/master/flexible-content-integration/src/main/scala/com/gu/flexiblecontent/integration/integration/cleaners.scala#L110).

@sihil because he said [this](https://github.com/guardian/flexible-content/issues/2483#issuecomment-211626214), @mchv because it’s a thing that (should) spans EVERYTHING!!!111110NE

mbarton:
https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringUtils.html#stripAccents(java.lang.String) maybe?

via https://stackoverflow.com/questions/3322152/is-there-a-way-to-get-rid-of-accents-and-convert-a-whole-string-to-regular-lette

Here's how Lucene does it, could add it as a dependency for a laff:

https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ASCIIFoldingFilter.java

currysoup:
Thanks both of you!

Yeah this is mostly just a getting started sort of thing. I'll implement the accent stripping and add rules for the characters Mat suggested.

akash1810:
this makes me feel queasy but in a good way 😂

PR #345, opened 2018-11-09
Stop calling Cancel – Reset

paperboyo:
Call things by what they do and by what ppl are used to.


paperboyo:
Yeah, Discard would work, too. Maybe even more unambiguous. Went with Cancel as it’s more common and (haven’t checked) should be more consistent with the rest of tooling.

PR #348, opened 2018-12-03
Check a reference is applicable to CAPI before issuing reindex

itsibitzi:
Diff external reference changes and make sure they're CAPI references before reindexing


itsibitzi:
I'm actually going to revert this because I'm fairly confident we can't trust that no one will ever rely on tags being indexed in flex even though  they're not used in content.

PR #352, opened 2019-01-29
Improve error handling of Tag image uploads

vlbee:
The reason for errors when uploading a tag image were not differentiated for users. 

This has now been fixed:

screen shot 2019-01-29 at 13 48 10
screen shot 2019-01-29 at 13 48 19
screen shot 2019-01-29 at 13 48 37


franziskas:
Is there a warning if the image is CMYK? This was the original problem that made us aware of the error message

itsibitzi:
We could potentially link out to a google doc which lists common issues. There is precedent for this in the contributor profile guidelines

PR #359, opened 2019-06-10
Blocking Ads/Contributions with tags

itsibitzi:
NB! Relies on: https://github.com/guardian/tags-thrift-schema/pull/26

Add two new fields on tags which allows composer to block tags or contributions asks.

Very simple on the frontend - just a little dropdown!

image

When a tag is added in Composer these fields are checked and Composer will take the appropriate action.


public/components/TagEdit/TagEdit.react.js
```diff
@@ -1,186 +1,243 @@
...
+    this.renderTagTypeSpecificFields.bind(this);
+  }
 
-      const sectionId = parseInt(e.target.value, 10);
-      const section = this.props.sections.filter((section) => section.id === sectionId)[0];
+  onUpdateSection(e) {
+    const sectionId = parseInt(e.target.value, 10);
+    const section = this.props.sections.filter(
+      section => section.id === sectionId
+    )[0];
```
akash1810: could use `.find` over `.filter` here to return a single item and avoid the `[0]`

public/components/TagEdit/TagEdit.react.js
```diff
@@ -1,186 +1,243 @@
...
-          </div>
+      })
+    );
+  }
+
+  renderTopicFields() {
+    return (
+      <div>
+        <div className="tag-edit__input-group" key="topic-section">
+          <label className="tag-edit__input-group__header">Section</label>
```
akash1810: unrelated to this PR, however we could add the [`for` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Attributes) to `label` to toggle the `input` on label click (bigger target area = better accessibility)

public/components/TagEdit/formComponents/TagAdBlocking.react.js
```diff
@@ -0,0 +1,60 @@
...
+  setContributionBlockingLevel(e) {
+    this.props.updateTag(
+      Object.assign({}, this.props.tag, {
+        contributionBlockingLevel: e.target.value
+      })
+    );
+  }
+
+  render() {
+    const blockingLevels = ["None", "Suggest", "Force"];
```
akash1810: Worth exporting these to DRY [this](https://github.com/guardian/tagmanager/pull/359/files#diff-ed9dd6f83909e3393f3af34d2a20747dR16) out a bit?

public/components/TagEdit/formComponents/TagAdBlocking.react.js
```diff
@@ -0,0 +1,60 @@
...
+            </option>
+          );
+        })}
+        )
+      </select>
+    );
+
+    return (
+      <div className="tag-edit__input-group">
+        <label className="tag-edit__input-group__header">Ad Blocking</label>
```
akash1810: minor: could this be `Block Ads` to more closely match `Block Contribution Asks`?

public/components/TagEdit/TagEdit.react.js
itsibitzi: Ah apologies, I left format on save on by accident so this file has a lot of noise in it :(

app/model/Tag.scala
```diff
@@ -1,47 +1,46 @@
...
+                isMicrosite: Boolean,
+                capiSectionId: Option[String] = None,
+                trackingInformation: Option[TrackingInformation],
+                campaignInformation: Option[CampaignInformation],
+                activeSponsorships: List[Long] = Nil,
+                sponsorship: Option[Long] = None, // for paid content tags, they have an associated sponsorship but it may not be active
+                paidContentInformation: Option[PaidContentInformation] = None,
+                expired: Boolean = false,
+                adBlockingLevel: Option[BlockingLevel],
+                contributionBlockingLevel: Option[BlockingLevel],
```
mkuzdowicz: Do `adBlockingLevel` needs to be an Option[] ? or maybe with BlockingLevel.None as default value ?
Just a thought

public/components/TagEdit/formComponents/TagAdBlocking.react.js
sihil: Also, the string values here are different to the string values in `BlockingLevel` which are all upper case.

PR #363, opened 2019-09-02
More logs for cache update

mkuzdowicz:
Introducing slightly more logs for TagLookupCache
It was not obvious before as if the tags were actually updated as TagEvent may not contain any tag as it is Option (coverd in logs) and the tag update function may ignore the tag update event

change includes as well adding unit tests for `insertTag` function

### DIFF
--- a/app/controllers/TagManagementApi.scala
+++ b/app/controllers/TagManagementApi.scala
@@ -1,19 +1,19 @@
 package controllers
 
-import com.amazonaws.services.pricing.model.FilterType
+import helpers.CORSable
 import model._
 import model.command.CommandError._
 import model.command._
+import model.forms.GetSpreadSheet
 import model.jobs.JobRunner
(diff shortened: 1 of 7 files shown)


app/controllers/TagManagementApi.scala
```diff
@@ -68,7 +68,11 @@ object TagManagementApi extends Controller with PanDomainAuthActions {
 
     val orderBy = req.getQueryString("orderBy").getOrElse("internalName")
 
+    val start = System.currentTimeMillis
     val tags = TagLookupCache.search(criteria)
+    val end = System.currentTimeMillis
+    Logger.debug(s"found tags: [${tags.map(_.internalName).mkString(", ")}] in TagLookupCache within ${end - start} milliseconds" +
```
akash1810: minor: can use multiline string here to avoid the `\n`

PR #365, opened 2019-09-03
FIX tags are not showing up in search results after creation

mkuzdowicz:
Fix for Tags not updating in local cache

- fixing serialisation of thrift Kinesis events
- relevant unit tests added


app/modules/clustersync/SectionSyncUpdateProcessor.scala
```diff
@@ -7,25 +7,40 @@ import play.api.Logger
...
+import scala.util.{Failure, Success, Try}
+
+object SectionEventDeserialiser {
 
   import services.KinesisRecordPayloadConversions._
 
-  override def process(record: Record) {
-    val sectionEvent = SectionEvent.decode(record)
+  def deserialise(record: Record): Try[SectionEvent] = {
+    val tProto = kinesisRecordAsThriftCompactProtocol(record, stripCompressionByte = true)
```
akash1810: Can we add a comment around why we're setting `stripCompressionByte` to true here?

app/modules/clustersync/TagSyncUpdateProcessor.scala
```diff
@@ -2,30 +2,53 @@ package modules.clustersync
...
 
-object TagSyncUpdateProcessor extends KinesisStreamRecordProcessor {
+import scala.util.{Failure, Success, Try}
+
+object TagEventDeserialiser {
 
   import services.KinesisRecordPayloadConversions._
 
+  def deserialise(record: Record): Try[TagEvent] = {
+    val tProto = kinesisRecordAsThriftCompactProtocol(record, stripCompressionByte = true)
```
akash1810: Can we add a comment around why we're setting `stripCompressionByte` to true here?

app/modules/clustersync/SectionSyncUpdateProcessor.scala
sihil: There is a test that checks this behaviour which kinda acts as a comment, but can't hurt.

app/modules/clustersync/TagSyncUpdateProcessor.scala
mkuzdowicz: sure

app/modules/clustersync/SectionSyncUpdateProcessor.scala
mkuzdowicz: yeah i can add that, test handles that but not say about context of this work-around as it is in some way a work-around

mkuzdowicz: added

app/modules/clustersync/TagSyncUpdateProcessor.scala
mkuzdowicz: added

PR #366, opened 2019-09-04
Adding test coverage for tag lookup cache and tag sync and fixing blocking level serialisation

mkuzdowicz:
# What was done here:
- Adding test coverage for tag lookup cache  
- Adding test coverage for tag sync
- fixing blocking level serialisation bug


test/repositories/TagLookupCacheTest.scala
```diff
@@ -0,0 +1,49 @@
...
+  override def beforeEach(): Unit = cache.allTags.set(Nil)
+
+  behavior of "TagLookupCache.insertTag()"
+
+  it should "insert new tag to cache" in {
+
+    val testTag = createTestTag("test", 1, 0)
+    cache.insertTag(testTag)
+
+    cache.allTags.get.contains(testTag) shouldEqual true
```
akash1810: Do we also want to assert about the cache state before insertion?

app/model/BlockingLevel.scala
```diff
@@ -1,26 +1,37 @@
...
-    }
+
+    def reads(json: JsValue): JsResult[BlockingLevel] = Reads.StringReads.reads(json).flatMap(v => {
+      val level = BlockingLevel.withName(v)
+      level match {
+        case BlockingLevel.NONE => JsSuccess(NONE)
+        case BlockingLevel.SUGGEST => JsSuccess(SUGGEST)
+        case BlockingLevel.FORCE => JsSuccess(FORCE)
+      }
+    })
```
sihil: Most of this boilerplate can be avoided by using the Play Framework extensions for enumeratum. See https://github.com/lloydmeta/enumeratum#play-integration

test/repositories/TagLookupCacheTest.scala
mkuzdowicz: sure i can add that

mkuzdowicz: done

app/model/BlockingLevel.scala
mkuzdowicz: including `"com.beachape" %% "enumeratum-play" % enumeratumPlayVersion` library introduce a lot of transitive dependencies conflicts like: ```Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:...``` fixing that may take some time, as health week finished already i will add this as a follow-up task: https://trello.com/c/eE3qw88c/864-tag-manager-use-enumeratum-play-library-to-avoid-boilerplate-on-enum-serialisation

PR #374, opened 2020-01-09
Allow people to filter for the existence of a field in the spreadsheet view

itsibitzi:
Fairly self explanatory. Allows you to filter by podcasts only (for example) or get all contributor tags with a byline image, etc.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "lodash.debounce": "^3.1.1",
     "lodash.startcase": "^4.4.0",
     "moment": "^2.10.6",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "lodash.debounce": "^3.1.1",
     "lodash.startcase": "^4.4.0",
     "moment": "^2.10.6",
-    "node-sass": "^3.3.3",
+    "node-sass": "4.13.0",
```
jonathonherbert: Sure thing! 🎉

PR #391, opened 2021-03-12
Revert "Move tagmanager to Graviton instances"

AWare:
Reverts guardian/tagmanager#389 rerevert at the end of the month. @srbd


jonathonherbert:
I don't think this is needed anymore, looks like we're already on ARM (see riffraff.yaml)

PR #404, opened 2021-05-18
upgrade fiesta

twrichards:
## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)


app/model/jobs/Job.scala
```diff
@@ -25,10 +30,13 @@ case class Job(
   createdAt: Long = new DateTime().getMillis // Created at in local time
 ) {
 
+  // FIXME is this OK??????
+  implicit val executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
twrichards, line 34: FYI @jonathonherbert

twrichards: managed to avoid this in the end

twrichards:
TESTED in CODE with Calvin (former CP) everything looks to be working now (even fixed a couple of bugs & typos along the way).... what a journey!

twrichards:
hey @itsibitzi - sorry for [excuse the pun] tagging you for review on this PR - but as one of the last remaining / most recent major contributors to this project I hope you can spare a little time and share any insight you can given the pretty broad and far-reaching nature of this PR 🙏

jonathonherbert (approved):
Reviewed commit by commit, nothing much to add – great work both. Really glad you've gone the extra mile to get this up to the current Play version.

Run locally and made a few edits as a smoke test and all looks good. Sounds like Calvin's taken a good look in CODE. @itsibitzi might have more to add for testing purposes. The only other thing I can think to test is that IIRC tag changes kick off tag reindexes in CAPI – might be worth testing this if it's not already been exercised.

Also nice to see a check for the node version in the startup script! 😁

twrichards:
Thanks for the review @jonathonherbert 

> The only other thing I can think to test is that IIRC tag changes kick off tag reindexes in CAPI – might be worth testing this if it's not already been exercised.

@JustinPinner @Fweddi wondering if you could help me confirm this. Are you familiar with the tag-manager CAPI interaction, I wonder if we could find a moment for you to watch the relevant parts of CAPI whilst we play around with tag-manager CODE?

JustinPinner:
I don't think this shouldn't pose a problem for us. If it runs as a reindex operation it's supposed to travel through our low priority stream, so it shouldn't cause any headaches for regular content updates. If you give us a shout when you're testing, we can look at what we can see coming through.

twrichards:
> I don't think this shouldn't pose a problem for us. If it runs as a reindex operation it's supposed to travel through our low priority stream, so it shouldn't cause any headaches for regular content updates. If you give us a shout when you're testing, we can look at what we can see coming through.

I've just looked in CAPI logs following doing a last bit of testing in CODE - things appear to show up so I'm going to go ahead and merge now 🙏

PR #387, opened 2021-01-18
chore: Bump node version

mxdvl:
## What does this change?

Uses the current active LTS is Fermium (v14.15.4).
Alternatively, it could be possible to use the maintenance LTS Dubnium (v12.20.1).

https://nodejs.org/en/about/releases/

## How to test & measure success?
The project should run locally.

## Have we considered potential risks?
Node v6 “Boron” was published [nearly 4 years ago][boron], and has [known security risks][].

[boron]: https://nodejs.org/en/blog/release/v6.9.0/
[known security risks]: https://www.cvedetails.com/vulnerability-list.php?vendor_id=12113&product_id=30764&version_id=203095&page=1&hasexp=0&opdos=0&opec=0&opov=0&opcsrf=0&opgpriv=0&opsqli=0&opxss=0&opdirt=0&opmemc=0&ophttprs=0&opbyp=0&opfileinc=0&opginf=0&cvssscoremin=0&cvssscoremax=0&year=0&month=0&cweid=0&order=3&trc=12&sha=cbc70f61ba78489641b25f8fedcbc90ab614d53f

PR #405, opened 2021-05-19
bump node to v14.17.0

mxdvl:
## What does this change?

Bumps node to a reasonable version, Fermium (LTS).
Redux of #387.

## How to test

Unit tests should pass. Run locally without issues.

More modern Node.

If node is running on a server, we need to make sure it works in that environment.


twrichards (approved):
Rebased now that we've fixed the build in https://github.com/guardian/tagmanager/pull/404

Tested in CODE - all seems fine 🤞  (note that node is only used used for build process - this isn't actually a node application)

PR #413, opened 2021-09-24
Update batch tagging interface

rhystmills:
## What does this change?

This PR updates the batch tagging interface to be a bit more user friendly.

It also fixes a small bug where the 'Removed' section of the status page actually lists tags to be added.

(N.B this PR previously included changes I thought would be necessary for the batch tagging fix in flexible-feed, but they weren't - hence all the commits).

**Before:**
Screenshot 2021-09-24 at 14 54 08

**After:**
Screenshot 2021-09-24 at 14 57 23

## How to test

Run tagmanager locally following the instructions [here](https://github.com/guardian/tagmanager), or deploy to CODE and test there.


SHession (commented):
Changes look sensible and the rationale makes sense.
I'm unfamiliar with the tool so more specific instructions on how to test locally would be helpful. It's also possible to clean up the git history if you believe most of it to be redundant.

rhystmills:
> Changes look sensible and the rationale makes sense. I'm unfamiliar with the tool so more specific instructions on how to test locally would be helpful. It's also possible to clean up the git history if you believe most of it to be redundant.

Cheers Sam - the best instructions for running the application are those on the [readme](https://github.com/guardian/tagmanager). 

In terms of testing this feature (which I now realise I haven't described very well):
1. Go to the 'Batch tag' tab in Tag Manager.
2. Search for an arbitrary article name, e.g. 'hello world', in 'Search by query'
3. Select the checkbox for one of those articles
4. Select a button from the bottom of the screen, e.g. 'Add tag to bottom of list'
5. Pick a tag from that list by searching and clicking the tag of choice (Optionally - repeat step 4 and 5 for one or more of the other buttons, i.e. Remove or Add to Top)
6. See and use the pop-up interface that should appear in the bottom right of the screen (which is affected by my changes)


Hope that's helpful!

SHession (approved):
Tested before / after, looks much better! 

Happy to pair if you want to tidying the commits, otherwise LGTM!

PR #418, opened 2021-11-26
Improve error handling for sponsorship image upload

rhystmills:
## What does this change?

This PR makes some improvements to error handling of sponsorship image uploads.

Specifically, it:
- Identifies images with incorrect file extensions (e.g. anything that isn't a .png or .jpg), and shows a user-friendly error message in this instance
- Displays the error `statusMessage` for errors that come from nginx instead of showing the html for an error page (as was previously the case). This includes files that are too large to be handled in a request (error `413`).
- Displays the error in a more distinct style (red text, error icon, comfortable margin).
- Models the validation functions as a for comprehension of `Either`s, as a representation of potentially failing dependent operations.

## How to test

Run `tagmanager` locally according to the [instructions on the readme](https://github.com/guardian/tagmanager) (with Composer credentials - run `npm install` then `./scripts/start.sh` if you've run the project before).

You will need additional permissions to access the commercial tags section, where you can create a sponsorship. You can [give these to yourself](https://permissions.code.dev-gutools.co.uk/admin) on CODE (also used for local development).

Try uploading bad file formats, files that are large, incorrect dimensions, correct dimensions.

#### This image should work:
Initech

## Images

**Incorrect file format:**
| Before | After |
| --- | --- |
|   image | image |

**File too large:**
* The error message here is derived from nginx - it's not particularly user-friendly, but I feel it's better than an html document - using nginx's error message should allow us to describe a greater range of http error statuses without writing specific messages for the less likely ones.

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -39,7 +39,6 @@ class Support(
 
   def checkFileExtension(image: File, filename: String): Either[String, File] = {
     val fileExtension = if(filename.contains(".")) filename.substring(filename.lastIndexOf(".")) else ""
-    }
 
     fileExtension match { 
       case ".jpg" => Right(image)
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -79,7 +79,7 @@ class Support(
     val requiredHeight = req.getQueryString("height").map(_.toLong)
 
     val imageValidationResult = for {
-      fileWithValidExtension <- checkFileExtension(picture.file, filename)
+      fileWithValidExtension <- checkFileExtension(picture.path.toFile, filename)
       image <- validateImageDimensions(fileWithValidExtension, requiredWidth, requiredHeight)
     } yield image
 
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -37,6 +37,20 @@ class Support(
 
   private val httpClient = new OkHttpClient.Builder().connectTimeout(5, TimeUnit.SECONDS).build
 
+  def checkFileExtension(image: File, filename: String): Either[String, File] = {
+    val fileExtension = filename.contains(".") match {
+      case true => filename.substring(filename.lastIndexOf("."));
+      case false => ""
+    }
+
+    fileExtension match { 
+      case ".jpg" => Right(image)
+      case ".jpeg" => Right(image)
+      case ".png" => Right(image)
+      case _ => Left("Image must have a file extension of '.jpg' or '.png'")
+    }
+  }
+
   def imageMetadata(imageUrl: String = "") = APIAuthAction {
      ImageMetadataService.fetch(imageUrl) match {
        case Right(imageMetadata) =>  Ok(Json.toJson(imageMetadata))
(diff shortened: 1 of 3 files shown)


app/controllers/Support.scala
```diff
@@ -37,6 +37,21 @@ class Support(
 
   private val httpClient = new OkHttpClient.Builder().connectTimeout(5, TimeUnit.SECONDS).build
 
+  def checkFileExtension(image: File, filename: String): Either[String, File] = {
+    val fileExtension = filename.contains(".") match {
+      case true => filename.substring(filename.lastIndexOf("."));
+      case false => ""
+    }
```
twrichards, line 44: [nit] pattern match is overkill for booleans, you can inline `if` in scala
```suggestion
    val fileExtension = if(filename.contains(".")) filename.substring(filename.lastIndexOf(".")) else ""
    }
```

app/controllers/Support.scala
```diff
@@ -45,30 +60,39 @@ class Support(
...
+  def validateImageDimensions(image: File, requiredWidth: Option[Long], requiredHeight: Option[Long]): Either[String, File] = {
+    val dimensionsAreValid = (requiredWidth, requiredHeight) match {
       case (None, None) => true
       case (Some(width), None) => width >= ImageMetadataService.imageWidth(image)
       case (None, Some(height)) => height >= ImageMetadataService.imageHeight(image)
       case (Some(width), Some(height)) => width >= ImageMetadataService.imageWidth(image) && height >= ImageMetadataService.imageHeight(image)
     }
+    dimensionsAreValid match {
+      case true => Right(image)
+      case false => Left(s"Image must have dimensions w:${requiredWidth.getOrElse("Not Specified")}px h:${requiredHeight.getOrElse("Not Specified")}px")
```
twrichards, line 72: another nit 
```suggestion
    if (dimensionsAreValid) Right(image) 
    else Left(s"Image must have dimensions w:${requiredWidth.getOrElse("Not Specified")}px h:${requiredHeight.getOrElse("Not Specified")}px")
```

app/controllers/Support.scala
```diff
@@ -45,30 +60,39 @@ class Support(
...
     val requiredHeight = req.getQueryString("height").map(_.toLong)
 
-    validateImageDimensions(picture.file, requiredWidth, requiredHeight) match {
-      case true => {
+    val imageValidationResult = for {
+      fileWithValidExtension <- checkFileExtension(picture.path.toFile, filename)
+      image <- validateImageDimensions(fileWithValidExtension, requiredWidth, requiredHeight)
+    } yield image
+
+    imageValidationResult match {
```
twrichards, line 86: it won't let me do a suggestion for the whole block, but I recommend using `.fold` on the Either to replace this outer pattern match, to avoid confusion with the inner pattern match on a different Either.

public/components/SponsorshipEdit/SponsorLogo.react.js
```diff
@@ -24,12 +24,19 @@ export default class SponsorEdit extends React.Component {
     this.props.onImageUpdated(undefined);
   }
 
+  constructErrorMessage(error){
+    if (error.status === 400) { // 400 errors are returned by our application, with useful responseText
+      return {errorMessage: error.responseText}
+    } else return {errorMessage: error.statusText} // Other errors are returned by nginx, where statusText is more useful
```
twrichards, line 30: best not mix with and without curlies...
```suggestion
    } else {
         return {errorMessage: error.statusText} // Other errors are returned by nginx, where statusText is more useful
    }
```

twrichards (approved):
brilliant work (I've only reviewed the code, not tested)
I've left a few comments/suggestions about code style (non are blocking)
👏

app/controllers/Support.scala
rhystmills: Good suggestion, have updated now.

app/controllers/Support.scala
rhystmills: Good suggestion, have implemented now.

PR #442, opened 2022-07-19
Upgrade to panda v1.0.6

andrew-nowak:
## What does this change?

Upgrades to pan-domain-auth v1.0.6 and panda-hmac v2.0.1

Also bumps scala & sbt versions and moves all sbt settings directly into build.sbt

PR #445, opened 2022-07-21
Include upgraded aws-java-sdk-sts

andrew-nowak:
## What does this change?

#442 included upgrades to AWS SDK modules, and changed to only include those that were required. The requirement for the STS module was missed, which meant that the resolved version was in 1.9.x (via kinesis-logstash-appender) causing runtime errors.

Explicitly depend upon the desired version of `aws-java-sdk-sts`


jonathonherbert (approved):
Just batch-tagged a thing on CODE, worked as intended 👍

PR #469, opened 2022-12-05
fix critical vlunerability in svg-sprite

andrew-nowak:
## What does this change?

Updates svg-sprite to avoid critical improper input validation vulnerability.

## How to test

The svg sprite file continues to be built.

No chance of malicious SVG files somehow crashing the build.


jonathonherbert (approved):
Tested in CODE, icons etc. look good to me.

(Build was failing as TC was running node 8, which does not like catch statements without parentheses, so I've bumped it to the version specified by the project.)

andrew-nowak:
thanks!

PR #480, opened 2023-02-10
[Snyk] Update scala dependencies

aracho1:
## What does this change?
Bump Scala dependencies which reduces the number of high Snyk vulnerabilities from 9 to 3

**before**
image
**after**
image

Tested on CODE and everything seems to run as expected.

## How to test
Deploy to CODE and check nothing breaks 

Reduce Snyk high vulnerabilities and the app still runs as it should

Tested on CODE and everything seems to run as expected


Fweddi (approved):
Compiled locally, can see it running in CODE 👍

PR #481, opened 2023-02-13
Remove GAPI script include

andrew-nowak:
## What does this change?

I can't see any evidence of the google api client being used in this repo (searches for `google` and `gapi` return nothing relevant) so delete the script tag that adds it to tagmanager.


phillipbarron (approved):
LGTM

PR #484, opened 2023-02-22
Add pocketcasts url

waisingyiu:
## What does this change?

MSS team are looking to provide Pocket Casts URL for podcast cards in MAPI so that the native app can allow users to open podcast with the Pocket Casts app directly.

This PR adds a pocketCastsUrl to tagmanager so that PocketCasts URL can be added to podcast tags.

Paired with @silvacb 

## How to test

We deployed the change to CODE and confirmed that the new UI field was added.

We tried setting the pocket casts URL in the new field and we were able to retrieve the URL when we opened the tag later.

## Images

Screenshot 2023-02-22 at 14 35 32


davidfurey (approved):
Great work!

PR #487, opened 2023-04-05
refactor Tag.asThrift to avoid getting from repository

andrew-nowak:
Builds on #485 to avoid the sponsorship lookup when creating the Thrift format of the Tag model, in cases where we already have the sponsorship data in scope (ie. when creating the tag)

Can't race against eventual consistency if you don't lookup from the database!


jonathonherbert:
@andrew-nowak didn't spot this when it was first created – still useful? Happy to give it a ✅ if re-opened.

(Was there a particular bug that prompted this change?)

PR #490, opened 2023-05-29
Add 5xx alarm

davidfurey:
DevX have recommended that we add an alert on 5xx errors. They have identified that occasionally this services returns a large number of 5xx response, and it would be good for us to be notified.

As it happens, this 5xx have always been related to bad requests being misreported as server errors (https://github.com/guardian/pan-domain-authentication/pull/118), but still probably worth adding an alert.

The approach is baed on targeting and media-service alerts.

Conveniently `/account/services/alert.topic.arn` is already set for this account.

### DIFF
--- a/cloudformation/tag-manager.yaml
+++ b/cloudformation/tag-manager.yaml
@@ -37,6 +37,10 @@ Parameters:
   CapiPreviewRole:
     Type: String
     Description: ARN of the CAPI preview role
+  AlertTopicArn:
+    Description: The ARN of the SNS topic to notify when an alarm occurs
+    Type: AWS::SSM::Parameter::Value<String>
+    Default: /account/services/alert.topic.arn
 Mappings:
   Config:
     CODE:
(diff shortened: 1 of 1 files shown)


cloudformation/tag-manager.yaml
```diff
@@ -303,3 +309,47 @@ Resources:
...
       Roles:
       - Ref: TagManagerRole
+  Tagmanager5XXAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: IsProd
+    Properties:
+      TreatMissingData: notBreaching
+      AlarmDescription: Tag manager is returning 5x responses
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: '0'
```
jacobwinch, line 319: I think you might want to raise this threshold slightly ([it's not unusual to see a handful of errors per hour](https://metrics.gutools.co.uk/goto/fd7u7XQ4z?orgId=1)) - perhaps 5 or 10 would be less noisy for the initial threshold?

jacobwinch (approved):

cloudformation/tag-manager.yaml
davidfurey: Good idea. Thanks for the review!

PR #492, opened 2023-06-02
bump pan-domain-auth and permissions

aracho1:
## What does this change?
Bump the pan-domain-auth to prevent invalid auth requests from throwing exceptions ([PR](https://github.com/guardian/pan-domain-authentication/pull/118)). The upgrade also required bumping the permissions library.

## How to test
Sending request without required cookies should raise bad requests instead of throwing exceptions.
Run Tag Manager on CODE. If you have the necessary permissions in CODE, you should be able to create and edit tags.

Invalid auth requests should stop producing noises in the logs and tag manager continues to function as normal.

Tested locally and on CODE

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -34,7 +34,7 @@ class AppComponents(context: Context, config: Config)
   // to request them. Seems to be a timing bug...
   //
   // This should only be a temporary fix @ 2016/02/09
-//  Permissions.list(PermissionsUser("preload@permissions"))
+//  Permissions.listPermission(PermissionsUser("preload@permissions"))
 
   val panDomainSettings = new PanDomainAuthSettingsRefresher(
     domain = config.pandaDomain,
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -1,4 +1,3 @@
-import com.gu.editorial.permissions.client.PermissionsUser
 import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
 import controllers.AssetsComponents
 import model.jobs.JobRunner
@@ -35,7 +34,7 @@ class AppComponents(context: Context, config: Config)
   // to request them. Seems to be a timing bug...
   //
   // This should only be a temporary fix @ 2016/02/09
-  Permissions.list(PermissionsUser("preload@permissions"))
+//  Permissions.list(PermissionsUser("preload@permissions"))
 
   val panDomainSettings = new PanDomainAuthSettingsRefresher(
     domain = config.pandaDomain,
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -1,4 +1,6 @@
+import com.amazonaws.auth.{AWSCredentialsProvider, DefaultAWSCredentialsProviderChain}
 import com.gu.pandomainauth.PanDomainAuthSettingsRefresher
+import com.gu.permissions.{PermissionsConfig, PermissionsProvider}
 import controllers.AssetsComponents
 import model.jobs.JobRunner
 import modules.clustersync.ClusterSynchronisation
@@ -12,6 +14,7 @@ import play.api.routing.Router
 import play.filters.HttpFiltersComponents
 import router.Routes
 import services._
+
 import scala.language.postfixOps
 
 class AppComponents(context: Context, config: Config)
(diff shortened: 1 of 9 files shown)


app/AppComponents.scala
```diff
@@ -35,7 +34,7 @@ class AppComponents(context: Context, config: Config)
   // to request them. Seems to be a timing bug...
   //
   // This should only be a temporary fix @ 2016/02/09
-  Permissions.list(PermissionsUser("preload@permissions"))
+//  Permissions.listPermission(PermissionsUser("preload@permissions"))
```
aracho1, line 37: I think this can go now as it seems to work for me but I'm not sure what the best way to test this is.

rhystmills: I'd be tempted to leave it in, given that it's been there for 7 years. It sounds like it could cause problems if a new permission is added, based on that comment, and its not causing any problems being there. 

I feel like the risk of the same obscure problem emerging if we add a permission in the future is higher than the value of removing it.

rhystmills: Origin PR doesn't have much on the Permissions issue it references: https://github.com/guardian/tagmanager/pull/149

rhystmills (approved):
Looks good to me, compiles and runs locally, and permissions are working in CODE (tested with create tag permission on and off) 👍

app/permissions/Permissions.scala
```diff
@@ -1,40 +1,32 @@
...
 
-  def getPermissionsForUser(email: String): Future[Map[String, Boolean]] = {
-    implicit val permissionsUser: PermissionsUser = PermissionsUser(email)
+  private val permissions: PermissionsProvider = PermissionsProvider(PermissionsConfig(Config().permissionsStage, Config().aws.region, credentials))
 
-    Permissions.list.map(_.filter(_._1.app == "tag-manager").flatMap( _ match {
-      case (p: Permission, PermissionGranted) => Map(p.name -> true)
-      case (p: Permission, PermissionDenied) => Map(p.name -> false)
-    }))
+  def testUser(email: String)(permission: PermissionDefinition): Boolean = {
```
jonathonherbert, line 26: Non-blocking: is there a reason we've swapped the curried arguments here? It might be quite useful to keep permissions as the first argument, as we can create instances of the function that are bound to particular permissions. (In the examples below I've removed the `Future` requirement per my other comment.)

For example, implementing `testAction` in `PermissionsActionCheck` is simpler:

```scala
val testAccess: String => Future[Boolean] =
    email => Future.successful(Permissions.testUser(email)(permission))
```

becomes

```scala
val testAccess: String => Boolean = Permissions.testUser(permission)
```

It might also help elsewhere – for example, in `SectionSpecificPermissions`, we have the line 

```scala
def commonTestAccess: PermissionDefinition => String => Future[Boolean] =
    permission => email => Future.successful(Permissions.testUser(email)(permission))
```

which could become

```scala
def commonTestAccess: PermissionDefinition => String => Boolean = Permissions.testUser
```

jonathonherbert (approved):
Tested in CODE, working as expected – great work! A few comments that are definitely not blocking.

app/permissions/Permissions.scala
aracho1: This was unintentional, thanks for spotting @jonathonherbert!

app/permissions/SectionSpecificPermissions.scala
```diff
@@ -17,9 +17,12 @@ object SectionPermissionMap {
 
 trait SectionSpecificPermissionActionFilter extends ActionFilter[UserRequest] {
 
-  val testAccess: (Permission => (String => Future[PermissionAuthorisation]))
+  val testAccess: (PermissionDefinition => (String => Future[Boolean]))
   val restrictedAction: String
 
+  def commonTestAccess: PermissionDefinition => String => Future[Boolean] =
+    permission => email => Future.successful(Permissions.testUser(email)(permission))
```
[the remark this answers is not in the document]
aracho1: Makes sense, thanks Jon!

app/AppComponents.scala
aracho1: Yes exactly, don't see this problem in other services that use the permissions library. Also in the original PR, this line of code was part of `PermissionsLoadModule`, which is now gone. I'm tempted to remove this, as leaving this in would require refactor due to the permissions version upgrade. I'll remove and test this in CODE.

PR #496, opened 2023-09-07
Add README instructions to connect to Path Manager via tunnel

samanthagottlieb:
## What does this change?

This adds instructions to the README on how to create a tunnel between your local server and the [Path Manager](https://github.com/guardian/path-manager) load balancer when running locally. This is needed to create tags, for example.

Previously the Path Manager URL was accessible via the VPN, however this is no longer the case, so instead we need to set up a tunnel to access this.

The DEV config in S3 has been amended, changing the path manager url to `localhost:6000` instead of the Path Manager URL - this is the port that should be used to create the tunnel.

## How to test

- follow instructions in the README to set up a tunnel
- run tag manager locally
- create a tag

If this works successfully, this indicates the tunnel has been created successfully and you have been able to access the required Path Manager URL to create a tag.


twrichards (approved):
great work 👏

PR #501, opened 2023-10-20
Bump React to version 16.14.0

rhystmills:
Co-authored-by: @jonathonherbert and @samanthagottlieb 

This PR bumps react in `tagmanager` to `16.14.0`, which will allow us to make use of React packages using hooks (introduced in `16.8.0`) - for example the upcoming [prosemirror-editor](https://github.com/guardian/prosemirror-editor/pull/1).

Bumping to `17`+ will be more complicated - some of our dependencies are still using the deprecated `componentWillMount`, `componentWillReceiveProps` and `componentWillUpdate` methods - which are removed in React 17 - leaving only their `UNSAFE_...` equivalents.

A few things that came up in the 15 > 16 React bump (may be relevant to atom-workshop and targeting if/when we do it there):
- The `componentWill…` and `componentDid…` lifecycle methods are now deprecated. To continue to use past React 16, use `UNSAFE_component…` instead. They're not 'unsafe' in terms of security, but apparently encouraged bad development practices: https://legacy.reactjs.org/blog/2018/03/27/update-on-async-rendering.html
-`prop-types` comes from its own library now. Import from there instead of using `React.PropTypes`
- the `moment` localiser has been moved to a separate package from version 3 > 4 (necessary to be compatible with React v16).
- `react-transition-group` needed a bump - meaning that `import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup'` became `import {CSSTransitionGroup} from 'react-transition-group';`

I've fixed a couple of pre-existing missing `key` prop warnings as well.

I did attempt bumping `react-router` but that one looks complicated as the API has changed - `browser-history` will come from a separate package, `Link` comes from a separate package (`react-router-dom`), `IndexRoute` no longer exists, and the app now needs to be wrapped in a `Router` or `BrowserRouter` among other snags.

## How to test

Run `tagmanager` locally according to the instructions in the readme, or deploy to CODE. Does the app run as expected? Do you see any React-themed errors?
(There will be some warnings about deprecated methods, but no errors)


dblatcher (approved):
LGTM

PR #502, opened 2023-11-23
support arbitrary urls for batch tagging

twrichards:
CP were asked to batch tag 1091 web urls with a tag and so it made sense to add that feature to the existing 'batch tag' functionality already present (which worked based on CAPI searches and ticking items). Fortunately the existing `batchTag` operation took simply a list of paths so most of this PR is just UI - encapsulated in a new React component `BatchTagArbitraryUrls`.

batch_arbitrary

### DIFF
--- a/public/components/BatchTag.react.js
+++ b/public/components/BatchTag.react.js
@@ -3,6 +3,7 @@ import ContentList from './ContentList/ContentList';
 import PageNavigator from './utils/PageNavigator.react';
 import BatchTagControls from './BatchTagControls/BatchTagControls';
 import BatchFilters from './BatchTag/BatchFilters.react';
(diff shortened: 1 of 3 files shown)


rhystmills:
Is it worth checking that each line is a valid URL before submitting them?
I tried adding a blank line into the pasted data and ended up with:
image
(So at least it fails in a controlled manner, but would be nice to spot this on submission.

twrichards:
> Is it worth checking that each line is a valid URL before submitting them? I tried adding a blank line into the pasted data and ended up with: image (So at least it fails in a controlled manner, but would be nice to spot this on submission. We could map the cleaned data down to only valid urls - or at least remove empty strings.

good spot @rhystmills - I've made it filter out empty (deployed to `CODE`)

rhystmills (approved):
Nice work, great to get a useful feature out so quickly!
Tested in CODE and working nicely.

PR #503, opened 2023-12-01
Deploy DynamoDB tables via Riff-Raff

jacobwinch:
## What does this change?

This PR allows Riff-Raff to deploy `tag-manager`'s DynamoDB tables. These are currently deployed manually via separate CloudFormation stacks[^1].

## How to test

WIP

It will be possible to apply all AWS infrastructure changes for this app via Riff-Raff.

Historically some teams were nervous about deploying databases using Riff-Raff, but we now have [protections in place](https://github.com/guardian/riff-raff/pull/623) to avoid the accidental deletion of stateful resource types. We deploy many DynamoDB tables across the estate using Riff-Raff and in some respects this reduces risk by avoiding the need to manually update production resources via the AWS Console.

[^1]: I think splitting this infrastructure across 3 separate CFN stacks is adding unnecessary complexity so it would be nice to combine them in the future. However, that is significantly more effort than just deploying all 3 using Riff-Raff so I don't plan to tackle that now.


riff-raff.yaml
```diff
@@ -20,3 +22,23 @@ deployments:
...
       templatePath: tag-manager.yaml
+
+  cloudformation-pillar-db:
+    type: cloud-formation
+    app: tag-manager-pillar-db
+    parameters:
+      templatePath: pillar-dynamo.yaml
+      prependStackToCloudFormationStackName: false
+      cloudFormationStackName: pillars-dynamo
+      cloudFormationStackByTags: false
```
jacobwinch, line 33: Generally I prefer tag based discovery, but this repo is already using stack name based discovery and I think it's better to remain consistent within the same repo

riff-raff.yaml
```diff
@@ -6,6 +6,8 @@ deployments:
     type: autoscaling
     dependencies:
       - cloudformation
+      - cloudformation-pillar-db
```
jacobwinch, line 9: We are going to deploy all 3 CFN stacks concurrently so it shouldn't slow down deployments too much

riff-raff.yaml
davidfurey: Fair enough

davidfurey (approved):
lgtm

PR #505, opened 2023-12-13
Use DevX logs instead of kinesis-logback-appender

davidfurey:
## What does this change?

This decouples the application from the logging, which avoids the application falling over if there is ever an issue pushing logging events to kinesis

Conveniently for this project the infrastructure definition is co-located with the app so there is not a separate editorial-tools-platform change.

For more context on why this approach is better than kinesis-logback-appender, see https://github.com/guardian/atom-workshop/pull/304

## How to test

Deploy this branch CODE check that logs still appear in central elk.


davidfurey:
Tested on CODE, logs still coming through

rhystmills (approved):

davidfurey:
Deployed to PROD, logs still coming through

PR #506, opened 2024-01-03
Fix 'user' on batch tag operations when viewed via Audit Logs

davidfurey:
## What does this change?

Batch tags were always logged as "default user" because a the logging function expected an implicit `User` object but the caller was expecting to provide an implicit `String` username. 

This issue highlighted how easy it is for bugs to creep in with implicit parameters that have default values, so I have also removed the default value from all functions that have an `implicit user(name): Option[String]` parameter. In doing so I identified a few places where we could be logging username but weren't.

The bug report also mentioned duplicated audit log entries. We were logging batch tag operations when the Job was created and when the Step was carried out. I have removed the Step level logging.

Tested in CODE. Old audit entry shows "Default User", newer entry shows david.furey

Screenshot 2024-01-03 at 12 26 00

## How to test

- [x] Deploy to CODE
- [x] Perform batch tag operation
- [x] Check that operation is only logged once
- [x] Check that operation is logged with your username rather than Default user

### DIFF
--- a/app/model/AppAudit.scala
+++ b/app/model/AppAudit.scala
@@ -34,15 +34,15 @@ object AppAudit extends Logging {
     }
   }
 
-  def reindexTags()(implicit username: Option[String] = None): AppAudit = {
+  def reindexTags()(implicit username: Option[String]): AppAudit = {
     AppAudit("reindexTags", new DateTime(), username.getOrElse("default user"), "tag reindex started");
   }
 
-  def reindexSections()(implicit username: Option[String] = None): AppAudit = {
+  def reindexSections()(implicit username: Option[String]): AppAudit = {
     AppAudit("reindexSections", new DateTime(), username.getOrElse("default user"), "section reindex started");
   }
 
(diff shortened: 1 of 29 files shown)


app/model/AppAudit.scala
```diff
@@ -34,15 +34,15 @@ object AppAudit extends Logging {
     }
   }
 
-  def reindexTags()(implicit username: Option[String] = None): AppAudit = {
+  def reindexTags()(implicit username: Option[String]): AppAudit = {
     AppAudit("reindexTags", new DateTime(), username.getOrElse("default user"), "tag reindex started");
```
twrichards, line 38: would `unknown` be better than `default user`?

twrichards (approved):
awesome - thanks for doing this @davidfurey . all the code changes LGTM and I've had a quick look at CODE where you'd already deployed this branch and things seem to look correct (improved) 👏

twrichards (approved):

PR #508, opened 2024-02-16
Sponsorship packages

davidfurey:
## What does this change?

Creates a new sponsorship attribute called `sponsorshipPackage`. This is to allow the US Ads team to sell sponsorships under the ‘Advertising partner’ or ‘Exclusive advertising partner’ labels. These labels will never be applied to Foundation funded journalism. 

An update to our externally facing content funding guidelines will be published before this change is used.

## How to test

Cannot be tested until tags-thrift-schema PR is released. See https://github.com/guardian/tags-thrift-schema/pull/43

* Deploy to CODE.
* ensure you have permission 
   image
* Attempt to create a new sponsored tag
* See that you can set the sponsorshipPackage to default, advertising partner or exclusive advertising partner
* See that once you have created the tag you can edit this field but cannot edit the sponsorship type
* Attempt to edit a pre-existing sponsored tag
* See that its sponsorshipPackage is set to the default value initially

US ads team are able to create new sponsorships that meet their requirements.

Existing tags will continue to have no value for this field until they are first edited. This is captured as an Option value in Scala and an optional field in thrift.

#### Before
Screenshot 2024-02-16 at 17 27 01

#### After
Screenshot 2024-02-16 at 17 26 37

#### Before
Screenshot 2024-02-16 at 17 38 59

#### After
Screenshot 2024-02-16 at 17 26 45

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/public/components/SponsorshipEdit/SponsorshipPackageEdit.react.js
+++ b/public/components/SponsorshipEdit/SponsorshipPackageEdit.react.js
@@ -1,0 +1,33 @@
+import React from 'react';
+
+export default class SponsorshipPackageEdit extends React.Component {
+
+    constructor(props) {
+        super(props);
+    }
+
+    updateType(e) {
+        this.props.updateSponsorship(Object.assign({}, this.props.sponsorship, {
+            sponsorshipPackage: e.target.value
+        }));
+    }
+
+
+    render () {
+
+        if (!this.props.sponsorship) {
+            return false;
+        }
+
+        return (
+            <div className="tag-edit__field">
+                <label className="tag-edit__label">Package</label>
+                <select value={this.props.sponsorship.sponsorshipPackage || "default"} onChange={this.updateType.bind(this)}>
+                    <option value="default">default</option>
+                    <option value="us">Advertising partner (US only)</option>
+                    <option value="us-exclusive">Exclusive advertising partner (US only)</option>
+                </select>
+            </div>
+        );
+    }
+}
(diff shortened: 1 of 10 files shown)


public/components/SponsorshipEdit/SponsorshipPackageEdit.react.js
```diff
@@ -0,0 +1,33 @@
...
+        this.props.updateSponsorship(Object.assign({}, this.props.sponsorship, {
+            sponsorshipPackage: e.target.value
+        }));
+    }
+
+
+    render () {
+
+        if (!this.props.sponsorship) {
+            return false;
```
twrichards, line 19: ~if you want it to render nothing, I think returning null is better to avoid `false` being printed to the screen~
```suggestion
            return null;
```

Well, TIL https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd

twrichards (approved):
tested in CODE - all working nicely AFAICT

Code LGTM and I learnt that `return false` from a component (as you have done) is better than `return null` when you don't want the component to render anything.

public/components/SponsorshipEdit/SponsorshipPackageEdit.react.js
davidfurey: So I should stick with `return false`? Most of this was copy+pasted from `SponsorshipPackageType.react.js`

PR #514, opened 2024-06-19
try to add description to spreadsheet tool

itsibitzi:
## What does this change?

Adds `description` field to spreadsheet search.

## How to test

Do a search on it


Fweddi (approved):

Fweddi (approved):

Fweddi:
After approving I noticed this for some descriptions - but might be a useful message:
image

itsibitzi:
`description` is an optional field that not all tags have. You can make the spreadsheet tool load an empty query which just returns everything, which I guess is what's happening here. It looks weird cos it doesn't relate to your query. I think this is a bug, separate to the PR that could be addressed when someone gets the time to properly investigate.

PR #515, opened 2024-07-08
Define DNS record in tagmanager CloudFormation

rhystmills:
## What does this change?

This PR adds a Guardian DNS Record custom CloudFormation resource to the stack, so that we can set the DNS record here, rather than manually managing it in our DNS provider.

The TTL is derived from the existing DNS records - the CloudFormation must exactly match the existing records in order for them to take effect.

For comparison:  a similar setup from another repo: https://github.com/guardian/acquisition-health-monitor/pull/9/files

## How to test

You can validate the yaml locally with:
`aws --region eu-west-1 --profile composer cloudformation validate-template --template-body file://cloudformation/tag-manager.yaml`

Deploying to CODE or PROD will validate the DNS record for that stage - if the deploy succeeds, the DNS record has been successfully added.

PR #516, opened 2024-07-08
Add CloudFront certificate

rhystmills:
Follow up to: https://github.com/guardian/tagmanager/pull/515

## What does this change?

This PR adds a CloudFront certificate to Tagmanager, to be used in a follow-up PR. Once merged, the ARN for the certificate will be added to Parameter Store and used as part of the CloudFormation for a CloudFront Distribution.

This requires its own RiffRaff project (`tag-manager-certificates`) and CI step because it must be in the `us-east-1` region, while the rest of the project is in `eu-west-1`.
 
## How to test

Deploy to CODE, as I have done [here](https://riffraff.gutools.co.uk/deployment/view/3e8bc850-74e2-46e8-9e4d-2a99079a2f57). If the deploy succeeds, the CloudFormation is valid.


davidfurey (approved):
lgtm

PR #517, opened 2024-07-10
Add Cloudfront Distribution to Tagmanager

rhystmills:
## What does this change?

This adds a CloudFront Distribution to Tagmanager, along with a CloudFront Cache Policy.

Traffic to Tagmanager will now go to via the Cloudfront CDN, rather than directly to our Load balancer, which should improve load times for users further from `eu-west-1` (Ireland), for example editorial staff in the US and Australia.

This branch is dependent on:
- https://github.com/guardian/tagmanager/pull/515 (define DNS record in CloudFormation).
- https://github.com/guardian/tagmanager/pull/516 (add Cloudformation Certificate)

In order for this PR to work in PROD, the Certificate PR will need to be merged, and the ARN of the created certificate in `us-east-1` will need to be manually added to Parameter Store in `eu-west-1` (e.g. in the AWS web dashboard), under the name `/PROD/flexible/tagmanager/cloudFrontCertificateArn`.

The PR also reduces the DNS times for PROD and CODE to a consistent `3600` seconds (i.e 1 hour)

## How to test

Deploy this branch to CODE. Does the deploy successfully complete?

Check back after at least 2 hours (the TTL set before this PR for the DNS record). Does the DNS record now point to the CloudFront distribution? 

You should be able to check the DNS record with `dig tagmanager.gutools.co.uk @dns1.p02.nsone.net.` and compare against the CloudFront DNS record in the AWS console.

### DIFF
--- a/cloudformation/tag-manager.yaml
+++ b/cloudformation/tag-manager.yaml
@@ -286,7 +286,7 @@ Resources:
     Type: "AWS::CloudFront::CachePolicy"
     Properties:
       CachePolicyConfig:
-        DefaultTTL: 3600
+        DefaultTTL: 0
         MaxTTL: 3600
         MinTTL: 0
         Name: !Sub "tag-manager-cache-policy-${Stage}"
--- a/cloudformation/tag-manager.yaml
+++ b/cloudformation/tag-manager.yaml
@@ -143,7 +143,7 @@ Resources:
           Action: sts:AssumeRole
           Resource: arn:aws:iam::642631414762:role/composerWriteToStaticBucket
       Roles:
-      - !Ref 'TagManagerRole'    
+      - !Ref 'TagManagerRole'
   TagManagerPermissionsBucketPolicy:
     Type: AWS::IAM::Policy
     Properties:
--- a/cloudformation/tag-manager.yaml
+++ b/cloudformation/tag-manager.yaml
@@ -59,14 +59,12 @@ Mappings:
       DesiredCapacity: 1
       InstanceType: t4g.medium
       DNSName: tagmanager.code.dev-gutools.co.uk
-      TTL: 7200
     PROD:
       MinSize: 3
       MaxSize: 6
       DesiredCapacity: 3
       InstanceType: t4g.medium
       DNSName: tagmanager.gutools.co.uk
-      TTL: 14400
 Resources:
   TagManagerRole:
     Type: AWS::IAM::Role
@@ -291,7 +289,7 @@ Resources:
         DefaultTTL: 86400
         MaxTTL: 31536000
         MinTTL: 0
-        Name: !Ref TagManagerCloudfrontCachePolicy
+        Name: !Sub "tag-manager-cache-policy-${Stage}"
         ParametersInCacheKeyAndForwardedToOrigin:
           CookiesConfig:
             CookieBehavior: "all"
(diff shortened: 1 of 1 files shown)


cloudformation/tag-manager.yaml
```diff
@@ -284,6 +282,69 @@ Resources:
         Value: flexible
       - Key: App
         Value: tag-manager
+  TagManagerCloudfrontCachePolicy:
+    Type: "AWS::CloudFront::CachePolicy"
+    Properties:
+      CachePolicyConfig:
+        DefaultTTL: 86400
```
davidfurey, line 289: Do you really want the default TTL to be 86400?

cloudformation/tag-manager.yaml
```diff
@@ -284,6 +282,69 @@ Resources:
...
+        MaxTTL: 31536000
+        MinTTL: 0
+        Name: !Sub "tag-manager-cache-policy-${Stage}"
+        ParametersInCacheKeyAndForwardedToOrigin:
+          CookiesConfig:
+            CookieBehavior: "all"
+          EnableAcceptEncodingBrotli: false
+          EnableAcceptEncodingGzip: false
```
davidfurey, line 297: Why not enable these?

cloudformation/tag-manager.yaml
```diff
@@ -284,6 +282,69 @@ Resources:
...
+            QueryStringBehavior: "all"
+  TagManagerCloudFrontDistribution:
+    Type: "AWS::CloudFront::Distribution"
+    Properties:
+      DistributionConfig:
+        Aliases:
+          - !FindInMap [Config, !Ref 'Stage', DNSName]
+        DefaultCacheBehavior:
+          AllowedMethods:
+            - "GET"
```
davidfurey, line 318: Looking at the routes file, I think you need to allow PUT, POST and DELETE too

cloudformation/tag-manager.yaml
```diff
@@ -284,6 +282,69 @@ Resources:
...
+            - "HEAD"
+            - "OPTIONS"
+          CachePolicyId: !Ref 'TagManagerCloudfrontCachePolicy'
+          Compress: true
+          TargetOriginId: !Ref 'TagManagerLoadBalancer'
+          ViewerProtocolPolicy: "allow-all"
```
davidfurey, line 324: I think this should be redirect-to-https

cloudformation/tag-manager.yaml
rhystmills: Let's make them true, I think I inherited these from another CFN and didn't have a reason to doubt them.

cloudformation/tag-manager.yaml
rhystmills: I borrowed those numbers from the presence snapshot and assumed they derived from gu/cdk for a sensible reason I didn't know - shall I make it 3600?

cloudformation/tag-manager.yaml
rhystmills: Will do!

davidfurey:
When testing on CODE, you should also try created and deleted a tag, and check that the interface isn't failing to update (due to caching)

cloudformation/tag-manager.yaml
rhystmills: Now set at 3600.

cloudformation/tag-manager.yaml
rhystmills: Now set to true.

cloudformation/tag-manager.yaml
rhystmills: Now done.

cloudformation/tag-manager.yaml
Fweddi: PATCH, too, I think.

Fweddi: > The parameter AllowedMethods is invalid (must be: [HEAD, GET] or [HEAD, GET, OPTIONS] or [HEAD, DELETE, POST, GET, OPTIONS, PUT, PATCH]).

rhystmills: Thanks Freddie, hadn't spotted that.

rhystmills:
> When testing on CODE, you should also try created and deleted a tag, and check that the interface isn't failing to update (due to caching)

Deployed to CODE and added then deleted a tag, all seemed to work quickly.

davidfurey (approved):
Approving but with 2 caveats:
1) when you tested in CODE did you wait for the TTL to expire?
2) it would be wise to change the TTL to 60s before merging this, because that will allow you to revert the change more quickly if there are problems. We can bump the TTL back up to 3600s in a follow up PR

rhystmills:
> Approving but with 2 caveats:
> 
> 1. when you tested in CODE did you wait for the TTL to expire?
> 2. it would be wise to change the TTL to 60s before merging this, because that will allow you to revert the change more quickly if there are problems. We can bump the TTL back up to 3600s in a follow up PR

That's now done.

Slightly embarrassingly, the first time I 'tested' my changes the deploy had failed and I hadn't noticed (CloudFront distributions have a `DomainName` and not a `DNSName` like a load balancer).

I've tested again, with enough time for the TTL to apply, and tag creation and deletion was working as expected.

rhystmills:
> > Approving but with 2 caveats:
> > 
> > 1. when you tested in CODE did you wait for the TTL to expire?
> > 2. it would be wise to change the TTL to 60s before merging this, because that will allow you to revert the change more quickly if there are problems. We can bump the TTL back up to 3600s in a follow up PR
> 
> That's now done.
> 
> Slightly embarrassingly, the first time I 'tested' my changes the deploy had failed and I hadn't noticed (CloudFront distributions have a `DomainName` and not a `DNSName` like a load balancer).
> 
> I've tested again, with enough time for the TTL to apply.
> 
> There do seem to be some issues with changes being reflected. Changed tag text doesn't seem to show straight away, and the most recently deleted tag seem to linger a bit longer in the search results.

Dropping the DefaultTTL to 0 has stopped the effects described above, and tag and search pages are working as before. I think we're good to merge now.

davidfurey (approved):
Great!

PR #518, opened 2024-07-10
enforce tag_manager_access permission

andrew-nowak:
## What does this change?

Follows up #513 to enforce the new tag_manager_access permission.

Need to grant access to all users who need to access before we can merge this...


bryophyta (approved):
lgtm!

PR #524, opened 2024-08-06
Remove Unwanted Resolvers

JamieB-gu:
Paired with @rtyley

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,8 +1,3 @@
-// The Typesafe repository
-resolvers += "Typesafe repository" at "https://repo.typesafe.com/typesafe/releases/"
-
-resolvers += "crappy twitter mvn repository" at "https://maven.twttr.com/"
-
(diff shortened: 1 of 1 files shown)


davidfurey (approved):

rtyley (approved):
Yep, we shouldn't be using additional resolvers that we don't need - checking them slows down the build, but also opens us up to additional places where we might download maliciously-corrupted artifacts!

The typesafe resolver itself is, as far as I'm aware, obsolete, and we only want to be loading our artifacts from Maven Central & OSS Sonatype.

See also:
* https://blog.oversecured.com/Introducing-MavenGate-a-supply-chain-attack-method-for-Java-and-Android-applications/
* https://docs.google.com/document/d/1B4TBmjsO8MfAo7kh-mwaEx-0yeGUrCmn5jvZi700yQc/edit?usp=sharing - DevX notes on MavenGate

PR #522, opened 2024-08-06
Pre-Scala 2.13: Upgrade dependencies to versions also available for Scala 2.13

rtyley:

PR #532, opened 2024-09-11
Pre-Scala 2.13: Upgrade deps to versions that are also available on Scala 2.13 (redux)

rtyley:
This is a repeat of the work in https://github.com/guardian/tagmanager/pull/522, previously reverted with https://github.com/guardian/tagmanager/pull/531 due to a runtime SLF4J error on startup (literally, running `sbt run` is enough to immediately expose this error, we should have done that!):

```
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J(W): Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J(W): Ignoring binding found at [jar:file:/usr/share/tag-manager/lib/ch.qos.logback.logback-classic-1.2.11.jar!/org>
SLF4J(W): See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
Oops, cannot start the server.
java.lang.ClassCastException: org.slf4j.helpers.NOPLoggerFactory cannot be cast to ch.qos.logback.classic.LoggerContext
```

* https://github.com/guardian/maintaining-scala-projects/issues/14

The error occurred because we were including an update to `content-api-client-default` that took it past version [v28.0.0](https://github.com/guardian/content-api-scala-client/releases/tag/v28.0.0), which upgraded `slf4j-api` to v2, which [no longer honours](https://www.slf4j.org/codes.html#ignoredBindings) the old static binder mechanism for finding logging providers. Logback v1.3+ copes with this, and is used by Play [v2.9](https://github.com/guardian/maintaining-scala-projects/issues/1). However, [Play 2.9 requires Scala 2.13](https://www.playframework.com/documentation/3.0.x/Migration29#Scala-2.12-support-discontinued) - and we are not yet on Scala 2.13! - so I'm going to just stick to CAPI client v27.0.0 for this PR.

### Original PR description

As preparation for Scala 2.13, this updates all dependencies so that they are at versions that are available for Scala 2.13 as well as Scala 2.12.

As ScalaTest is updated to ScalaTest v3.2, we had to do some code updates to cope with class/package changes. We did these manually, though later we remembered that maybe we could have used Scalafix, as in this example here:

* https://github.com/guardian/content-api/pull/2893

The steps would have been:

- add sbt-scalafix plugin to user-level sbt plugins file
  - addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.12.1")
- checked out main branch, still scala 2.12 (and so already compiled successfully)
- launch sbt
- run `scalafixEnable`
- run `scalafixAll dependency:RewriteDeprecatedNames@org.scalatest:autofix:3.1.0.1`
- increment scalatest dependency

## Testing

This has been [successfully deployed](https://riffraff.gutools.co.uk/deployment/view/d4cd0773-4c5b-4a74-a8c6-9cf24aad8248) to https://tagmanager.code.dev-gutools.co.uk/ (which we didn't do with https://github.com/guardian/tagmanager/pull/522!) and the tool looks good:

image

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "tag-manager"
 
 version := "1.0"
 
-lazy val scalaVer = "2.12.16"
+lazy val scalaVer = "2.12.19"
 
 resolvers ++= Resolver.sonatypeOssRepos("releases")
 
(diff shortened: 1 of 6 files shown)


JamieB-gu (approved):

rtyley:
This has successfully [deployed](https://riffraff.gutools.co.uk/deployment/view/49692c96-b184-4f92-87e0-4ef70bb296b0) to PROD, and I've verified that I can still re-authenticate.

PR #536, opened 2024-10-11
Upgrade to Scala 2.13

rtyley:
* https://github.com/guardian/maintaining-scala-projects/issues/2

This PR sits on top of:

* https://github.com/guardian/tagmanager/pull/522
* https://github.com/guardian/tagmanager/pull/523
* https://github.com/guardian/tagmanager/pull/526
* https://github.com/guardian/tagmanager/pull/532
* https://github.com/guardian/tagmanager/pull/538

...with all those updates out of the way, this PR to upgrade to Scala 2.13 is pretty small:

* Use https://github.com/rtyley/scala-collection-plus to get `mapV` to replace `mapValues`
* [`collection.breakOut`](https://www.scala-lang.org/api/2.12.3/scala/collection/index.html#breakOut[From,T,To](implicitb:scala.collection.generic.CanBuildFrom[Nothing,T,To]):scala.collection.generic.CanBuildFrom[From,T,To]) no longer exists - it was just a way of converting a collection to particular type, and we can just say `toMap` here instead.
* `scala.collection.convert.wrapAll._` goes to  `scala.jdk.CollectionConverters._`

Completing this upgrade is also a step in unblocking:

* https://github.com/guardian/pan-domain-authentication/issues/149

## Testing

This has been [deployed](https://riffraff.gutools.co.uk/deployment/view/bf342770-237c-4c67-b691-32d0926ba7a9) to CODE, and looks good:

image

PR #538, opened 2024-10-22
Scala 2.13 Scalafix: `ExplicitResultTypes` for implicit definitions

rtyley:
This is a precursor to:

* https://github.com/guardian/tagmanager/pull/536

Scala 2.13.11 added a warning that implicit definitions [should]( https://nrinaudo.github.io/scala-best-practices/tricky_behaviours/type_implicits.html) have an explicit type (because implicit resolution is already complicated enough for the compiler, and things like file-order can actually make a difference to whether implicit scopes are correctly searched: https://github.com/scala/bug/issues/8697#issuecomment-292432445):

* https://github.com/scala/scala/pull/10083

The warning looks like this:

```
[error] ~/code/presence-indicator/app/autoscaling/Notification.scala:7:16: Implicit definition should have explicit type (inferred play.api.libs.json.Reads[autoscaling.Notification]) [quickfixable]
[error]   implicit val jsonReads = Json.reads[Notification]
[error]                ^
```

...this prepares us for Scala 3, where the explicit type is _required_.

More widely, beyond implicit definitions, in **library** code, the best practice is to always add an explicit type to all your **public** members, even when you're happy with what's being inferred - otherwise you can unintentionally break binary-compatibility just by changing the _implementation_ of a field:

* https://docs.scala-lang.org/scala3/guides/migration/incompat-type-inference.html
* https://nrinaudo.github.io/scala-best-practices/binary_compat/explicit_type_annotations.html
* https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html

## Automatically fixing this code issue

Scalafix actually does a better job than `-quickfix` for this particular task, because it adds imports if it needs to, so that you end up with this in your code:

```
implicit val jsonReads Reads[Notification] = Json.reads[Notification]
```

...rather than something like:

```
implicit val jsonReads Reads[com.gu.blah.foo.bar.Notification] = Json.reads[Notification]
```

### Fixing while still on Scala 2.12 - use Scalafix

In this commit, we're only trying to fix the implicit definitions, so I've added this in a new `.scalafix.conf` config file:

```
ExplicitResultTypes.onlyImplicits = true
```

The Scalafix rule needs to be run while the project is still on Scala 2.12, not Scala 2.13 (otherwise sbt will say: "Error downloading ch.epfl.scala:sbt-scalafix;sbtVersion=1.0;scalaVersion=2.13:0.13.0").

Once the Scalafix plugin is made available to sbt (by adding `addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.13.0")` to either `project/plugins.sbt` or `~/.sbt/1.0/plugins.sbt`), you can run these commands on the sbt prompt to automatically generate the changes in this PR:

```
scalafixEnable
scalafixAll ExplicitResultTypes
```

### DIFF
--- a/app/model/CampaignInformation.scala
+++ b/app/model/CampaignInformation.scala
@@ -18,7 +18,7 @@ case class CampaignInformation(campaignType: String) {
 
 object CampaignInformation {
 
-  implicit val trackingFormat = Jsonx.formatCaseClass[CampaignInformation]
+  implicit val trackingFormat: OFormat[CampaignInformation] = Jsonx.formatCaseClass[CampaignInformation]
 
   def apply(thriftCampaignInformation: ThriftCampaignInformation): CampaignInformation =
     CampaignInformation(
--- a/app/model/ClientConfig.scala
+++ b/app/model/ClientConfig.scala
@@ -2,6 +2,7 @@ package model
 
 import ai.x.play.json.Encoders.encoder
 import ai.x.play.json.Jsonx
+import play.api.libs.json.OFormat
 
 case class ClientConfig(username: String,
                         capiUrl: String,
@@ -15,5 +16,5 @@ case class ClientConfig(username: String,
                        )
 
 object ClientConfig {
-  implicit val clientConfigFormat = Jsonx.formatCaseClass[ClientConfig]
+  implicit val clientConfigFormat: OFormat[ClientConfig] = Jsonx.formatCaseClass[ClientConfig]
 }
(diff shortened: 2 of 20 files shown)


JamieB-gu (approved):

rtyley:
[Deployed](https://riffraff.gutools.co.uk/deployment/view/38d5c592-070e-4b9b-94fa-f71723203b88) to PROD, looks good!

image

PR #547, opened 2025-02-27
Panda: monitor non-active-key use

rtyley:
See:

* https://github.com/guardian/pan-domain-authentication/pull/182
* https://github.com/guardian/pan-domain-authentication/releases/tag/v9.0.0

With this change, the [logs](https://logs.gutools.co.uk/app/r/s/YWdfP) tell us when a _non-active key_ is used:

image

Example log message:

> NON_ACTIVE_KEY_USED (used='1xpqc', active='6k065'): active_freq=Map(false -> 22, true -> 3) key_freq=Map('1xpqc' -> 25)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "tag-manager"
 
 version := "1.0"
 
-lazy val scalaVer = "2.13.15"
+lazy val scalaVer = "2.13.16"
 
 resolvers ++= Resolver.sonatypeOssRepos("releases")
 
@@ -19,7 +19,7 @@ scalacOptions ++= Seq(
 
 lazy val awsVersion = "1.12.403"
 
-val pandaVersion = "7.0.0"
+val pandaVersion = "9.0.0"
 lazy val dependencies = Seq(
   "com.amazonaws" % "aws-java-sdk-dynamodb" % awsVersion,
   "com.amazonaws" % "aws-java-sdk-ec2" % awsVersion,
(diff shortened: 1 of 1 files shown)


bryophyta (approved):
we tested a preview release of Panda 9 on CODE and it worked as expected 👍

abeddow91 (approved):

build.sbt
```diff
@@ -4,7 +4,7 @@ name := "tag-manager"
 
 version := "1.0"
 
-lazy val scalaVer = "2.13.15"
+lazy val scalaVer = "2.13.16"
```
rtyley, line 7: This PR requires the upgrade to Scala v2.13.16 for this chain of reasons:

* Panda v9 is compiled with Scala v2.13.16 - but in the past, that _wouldn't_ have required code using the library (like `tagmanager` to be updated to that Scala version
* [Scala Improvement Proposal (SIP) 51](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html) was a proposal created in December 2022 to _"Drop Forwards Binary Compatibility of the Scala 2.13 Standard Library"_ - the proposal explained that this would [make improving the Scala standard library easier](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html#motivation), and was approved.
* sbt [v1.10.0](https://github.com/sbt/sbt/releases/tag/v1.10.0), released May 2024, was updated to respect this: a project's `scalaVersion` can _no longer_ be a lower 2.13.x version number than its transitive dependencies

This project uses sbt 1.10, so this new rule takes effect:

https://github.com/guardian/tagmanager/blob/84a6157749c910749342784beac5754634d036c4/project/build.properties#L1

PR #552, opened 2025-04-15
Fix exclude tag filter

davidfurey:
## What does this change?

Fixes an operator precedence bug which was causing e.g. JDVance topic tag to be constructed as "undefinedjd-vance" instead of "us-news/jd-vance"

## How to test

Deploy to code. Attempt to exclude jd vance. See that tag appears correctly.


Fweddi (approved):
Seems reasonable

davidfurey:
Works in CODE

PR #588, opened 2025-12-11
Revert "Add keyword type"

simonbyford:
Reverts guardian/tagmanager#567

Comms need to go out first, and data backfilled


JustinPinner (approved):
As a straight up revert this should be safe. Caveat: I haven't been involved in any of these changes so cannot reasonably judge any potential side effects. Please monitor carefully after the reversion is deployed.

PR #593, opened 2026-01-08
Add keyword type backfill

simonbyford:
## What does this change?

Adds a mechanism to backfill the new `keywordType` field for existing topic tags.

The data science team have provided us with a csv file containing a list of tag paths (e.g. `film/inside-out`) mapped to their keyword type (e.g. `WORK_OF_ART_OR_PRODUCT`). Here is a sample to illustrate the format:

```
id,webTitle,internalName,keywordType
education/ravensbournecollege,Ravensbourne,Ravensbourne College,PLACE
sport/nhl,NHL,NHL,ORGANISATION
sport/nick-kyrgios,Nick Kyrgios,Nick Kyrgios,PERSON
sport/olympic-torch,Olympic torch,Olympic torch,OTHER
stage/a-taste-of-honey,A Taste of Honey,A Taste of Honey,WORK_OF_ART_OR_PRODUCT
stage/adelaide-fringe-2015,Adelaide fringe 2015,Adelaide fringe 2015,EVENT
```

This PR introduces a new `/migration/keywordTypes` route which provides an interface to upload this csv file and process its contents.

When processing begins, the system iterates through all the mappings and updates each tag's `keywordType` field accordingly. It also publishes updates to the Kinesis stream (to update CAPI) and creates an audit record for each update.

If a given tag cannot be found, or its associated keyword type is malformed, the row is skipped but processing continues.

### Happy path

1) Deploy to CODE and visit: https://tagmanager.code.dev-gutools.co.uk/migration/keywordTypes
2) Upload a csv file containing a small handful of mappings in the appropriate format. The above example should work.
3) Once the job is started, visit the job status page to see its progress: https://tagmanager.code.dev-gutools.co.uk/status
4) When complete, find the appropriate tags and ensure:
- their keyword type has been set correctly ([example](https://tagmanager.code.dev-gutools.co.uk/tag/64543))
- an audit record has been created ([example](https://tagmanager.code.dev-gutools.co.uk/tag/64543))
- CAPI is up to date ([example](https://content.code.dev-guardianapis.com/tags?q=nick-kyrgios&api-key=))

Finally, [check the logs](https://logs.gutools.co.uk/s/editorial-tools/app/r/s/DvJl1) to see if there are any errors. 

### Non-existent keyword type

Try with a non-existent keyword type. For example:

```
sport/nick-kyrgios,Nick Kyrgios,Nick Kyrgios,BROKEN
```

This should result in the mapping being skipped and an error logged:

```
[...]

PR #597, opened 2026-01-21
Fix csv parsing bug

simonbyford:
## What does this change?

Follows on from https://github.com/guardian/tagmanager/pull/593

There was a bug in the CSV parser which meant that certain rows were not being processed correctly.

For example, this row would be processed fine:

```
books/colin-barrett,Colin Barrett,Colin Barrett,PERSON
```

Whereas:

```
books/colette,Colette,"Colette (French author, d.1954)",PERSON
```

Would fail with the error:

```
Invalid keyword type 'd.1954)"' for tag 'books/colette', skipping
```

This is because the tag name contains a comma, and the quotes surrounding the tag name aren't being respected.

Using a proper CSV parsing library instead of relying on `.split(",")` should fix the issue.

## How to test

Tested on CODE, everything looks good


jonathonherbert (approved):
👍  sorry to miss in the original PR!

PR #626, opened 2026-04-01
Upgrade to React 17

davidfurey:
## Upgrade to React 17

Almost entirely AI

Upgrades the frontend from React 16 to React 17.0.2 and migrates the full ecosystem of React-related dependencies to compatible versions.

### Why

React 16 is end-of-life. Several transitive dependencies (notably `react-router` v3) rely on the legacy React context API. This upgrade brings us to React 17, eliminates deprecated APIs, and keeps the dependency tree supportable.

React 19 is the eventual target but is currently blocked on `@guardian/prosemirror-editor` publishing a compatible release.

### What changed

**Package upgrades**

| Package | Before | After |
|---|---|---|
| `react` / `react-dom` | 16 | 17.0.2 |
| `react-router` | 3 (removed) | `react-router-dom` 6 |
| `react-redux` | 5 | 8 |
| `redux` | 4 | 5 |
| `redux-thunk` | 1 | 3 |
| `redux-logger` | 2 | 3 |
| `react-transition-group` | 1 | 4 |
| `react-dropzone` | 3 | 14 |
| `react-widgets` | 4 | 5 |
| `react-tooltip` | 4 | 5 |
| `react-hot-loader` | removed (abandoned) | — |

**Entry point (`app.js`)** — Updated to use `ReactDOM.render` with the new `RouterProvider`.

**Router (`router.js`)** — Full rewrite from react-router v3's JSX `<Router>`/`<Route>` tree to react-router-dom v6's `createBrowserRouter` data-object format:
- `onEnter` permission guards replaced with a `<ProtectedRoute>` component
- `onLeave={clearErrorBar}` moved into `componentWillUnmount` on `TagDisplay`
- Route-injected props (`routeParams`, `isMicrositeView`) are now passed explicitly via HOC (`withParams`) and element props respectively
- Telemetry `onUpdate` callback moved to a `useLocation`-based hook in `ReactApp`

**`ReactApp`** — `this.props.children` replaced with `<Outlet />`.

**`Header`** — `<Link activeClassName>` replaced with `<NavLink className={({ isActive }) => ...}>`.

**`react-transition-group`** — `CSSTransitionGroup` (removed in v2) replaced with `TransitionGroup` + `CSSTransition` in 3 components.

**`react-dropzone`** — v3 JSX `<Dropzone>` component replaced with the v14 `useDropzone` hook via a small functional wrapper in `SponsorLogo`.

**`react-widgets`** — Import paths updated, `DateTimePicker` renamed to `DatePicker` (removed in v5), CSS import updated, `react-widgets-moment` localizer removed (now built-in) in 3 components.

**`react-tooltip`** — `data-tip` → `data-tooltip-id`/`data-tooltip-html`, `<ReactTooltip>` → `<Tooltip id="...">` in 5 components.

**`redux-thunk` / `redux-logger`** — Updated to named exports (`{ thunk }`, `{ createLogger }`).

### Testing

- Production webpack build passes with no errors (only pre-existing bundle-size warnings)
- Smoke test: navigate each route (tag search/edit, section, sponsorship, audit, batch tag, merge tag, mapping), verify active nav link highlights, date pickers render, image dropzone works

### What's not in scope

Class components have been left as-is — they remain valid in React 17. The `UNSAFE_` lifecycle methods present in several components (`UNSAFE_componentWillReceiveProps`, `UNSAFE_componentWillMount`) still work and are not removed here.


public/components/BatchTagControls/BatchTagControls.js
```diff
@@ -173,14 +173,16 @@ export default class BatchTagControls extends React.Component {
 
       const {toAddToTop, toAddToBottom, toRemove} = this.state;
       return (
-        <CSSTransitionGroup transitionName="batch-status-transition" transitionEnterTimeout={500} transitionLeaveTimeout={500}>
-          <div className="batch-status__container">
-            <div className="batch-status">
-              {this.renderMode()}
+        <TransitionGroup>
```
abeddow91, line 176: This works fine but I don't think this TransitionGroup wrapper is necessary. From what I understand it is only being used here for one transition but the docs suggest its use case is for managing `a set of transition components (<Transition> and <CSSTransition>) in a list` (https://reactcommunity.org/react-transition-group/transition-group )

Given the scope of this PR is reasonable this could be tidied up afterwards

public/components/Tag/Display.react.js
```diff
@@ -34,6 +36,10 @@ class TagDisplay extends React.Component {
       }
     }
 
+    componentWillUnmount() {
+      getStore().dispatch(clearError());
+    }
+
```
abeddow91, line 41: This seems to work fine but It seems a bit odd that we're calling the store directly here rather than calling it through props as per the rest of the file. It might be more consistent to have this wired up through mapDispatchToProps.

public/components/Tag/TagValidation.react.js
```diff
@@ -23,9 +23,13 @@ export default class TagValidation extends React.Component {
     return (
       <div className="tag-validation">
         <div className="tag-validation__header">Validation</div>
-        <CSSTransitionGroup transitionName="validation-transition" transitionEnter={false} transitionLeaveTimeout={500}>
-          {this.props.validations.map(this.renderValidationResult, this)}
-        </CSSTransitionGroup>
+        <TransitionGroup>
+          {this.props.validations.map(validation => 
+            <CSSTransition key={validation.message}classNames="validation-transition" timeout={{ enter: 0, exit: 500 }}>
```
abeddow91:
```suggestion
            <CSSTransition key={validation.message} classNames="validation-transition" timeout={{ enter: 0, exit: 500 }}>
```

public/components/utils/SaveButton.react.js
```diff
@@ -41,9 +37,13 @@ export default class SaveButton extends React.Component {
 
     render () {
       return (
-        <CSSTransitionGroup transitionName="save-transition" transitionEnterTimeout={500} transitionLeaveTimeout={500}>
-          {this.renderButtons()}
-        </CSSTransitionGroup>
+      <TransitionGroup>
```
abeddow91, line 40: Same as before, I don't think this wrapper is needed as we only have one csstransition, not a dynamic list

public/router.js
```diff
@@ -1,82 +1,73 @@
...
+const SponsorshipDisplayWithParams = withParams(SponsorshipDisplay);
 
-function onRouterUpdate() {
-  const domainMatch = /^.*\.(?<environment>local|code)\.dev-gutools\.co\.uk$|^.*\.gutools\.co\.uk$/
-      .exec(location.hostname);
-  if (domainMatch) {
-    const stage = (domainMatch.groups?.environment || "PROD");
-    const telemetryUrl = stage === "PROD" ? "user-telemetry.gutools.co.uk" : `user-telemetry.${stage}.dev-gutools.co.uk`
-    const image = new Image();
-    image.src = `https://${telemetryUrl}/guardian-tool-accessed?app=tag-manager&stage=${stage.toUpperCase()}&path=${window.location.pathname}`;
```
abeddow91, line 51: This has been moved into ReactApp.js

abeddow91 (approved):
Tested on CODE and all working as expected. I've left a few comments but given the scope of this PR I think its reasonable that they could be follow ups if desired, they are not blocking changes.

public/components/BatchTagControls/BatchTagControls.js
davidfurey: Great point! I was trying to remove the CSSTransition element but removing the TransitionGroup is the way to go. Done.

public/components/Tag/Display.react.js
davidfurey: I do like consistency, so I've attempted the change you've suggested. Let me know if you think it is okay

public/components/BatchTagControls/BatchTagControls.js
davidfurey: Actually, the animation doesn't happen without the TransitionGroup, so I've kept it

public/components/Tag/Display.react.js
abeddow91: Spot on! thank you 👏

abeddow91 (approved):

PR #628, opened 2026-04-01
Bump kinesis (again) 

Georges-GNM:
Reverts guardian/tagmanager#627 and should fix the issue seen in prod. The second commit in this PR contains the fix. 

Although I'd tested the original PR in code and it seemed to be working as intended, in prod relevant stream updates weren't coming through. We can confirm this by looking at the logs:
- a tag (local elections 2026) created prior to the original PR getting released provides [these logs](https://logs.gutools.co.uk/s/editorial-tools/app/r/s/noBAr)
- a tag (april fools day) created after shows [these](https://logs.gutools.co.uk/s/editorial-tools/app/r/s/mHeVS).

The salient difference is that the `TagEvent received` message never appears in the latter, and on double checking, weren't coming through when the branch was up on code either ([searching](https://logs.gutools.co.uk/app/r/s/8YnPO) for it in the past couple of weeks only brings up records from today). 

## What's changed

To be transparent, I used Copilot to help investigate this, and it pointed to an issue in the ByteBuffer handling inside KinesisRecordPayloadConversions.

When consuming Kinesis records via the updated KCL client, KinesisClientRecord.data() can return ByteBuffer instances with a non-zero arrayOffset or position, due to KCL's internal pooled/sliced buffer allocation. The previous implementation didn't account for this - it passed the buffer directly without respecting its current position, which meant Thrift could be decoding from the wrong byte index, causing deserialization failures.

The fix updates kinesisRecordAsThriftCompactProtocol to correctly extract the relevant bytes from the ByteBuffer by respecting its position and arrayOffset, so the correct slice of the backing array is always read regardless of how the buffer was constructed.

Additional tests have been added to TagEventDeserialiserTest covering the three ByteBuffer states KCL v2 can produce: clean, sliced (arrayOffset > 0), and positioned (position > 0).

## Testing

Deployed to CODE, I created a new tag (gl-test-kinesis) and in quick succession updated it a few times. These updates came through correctly, and we can see the `TagEvent received` messages in the [logs](https://logs.gutools.co.uk/app/r/s/OqVXn). Considering the level of usage of the code environment vs prod, it feels like there's a limit to how well this can be tested, but those logs give me reassurance that the kinesis stream logic is now correct.

### DIFF
--- a/test/modules/clustersync/TagEventDeserialiserTest.scala
+++ b/test/modules/clustersync/TagEventDeserialiserTest.scala
@@ -1,6 +1,6 @@
 package modules.clustersync
 
-import com.amazonaws.services.kinesis.model.Record
+import software.amazon.kinesis.retrieval.KinesisClientRecord
 import com.gu.tagmanagement.{EventType, TagEvent}
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
@@ -17,7 +17,9 @@ class TagEventDeserialiserTest extends AnyFlatSpec with Matchers  {
 
     val thriftKinesisEvent: Array[Byte] = ThriftSerializer.serializeToBytes(tagUpdateEvent, true)
 
-    val testRecordWithCompression = new Record().withData(ByteBuffer.wrap(thriftKinesisEvent))
+    val testRecordWithCompression = KinesisClientRecord.builder()
+      .data(ByteBuffer.wrap(thriftKinesisEvent))
+      .build()
 
     TagEventDeserialiser.deserialise(testRecordWithCompression) shouldBe Success(tagUpdateEvent)
 
--- a/test/modules/clustersync/TagEventDeserialiserTest.scala
+++ b/test/modules/clustersync/TagEventDeserialiserTest.scala
@@ -9,20 +9,37 @@ import services.ThriftSerializer
 import java.nio.ByteBuffer
 import scala.util.Success
 
-class TagEventDeserialiserTest extends AnyFlatSpec with Matchers  {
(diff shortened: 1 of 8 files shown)


test/modules/clustersync/TagEventDeserialiserTest.scala
```diff
@@ -9,20 +9,37 @@ import services.ThriftSerializer
...
       .data(ByteBuffer.wrap(thriftKinesisEvent))
       .build()
 
     TagEventDeserialiser.deserialise(testRecordWithCompression) shouldBe Success(tagUpdateEvent)
+  }
+
+  it should "deserialise a sliced ByteBuffer (non-zero arrayOffset, as produced by KCL v2 buffer pooling in production)" in {
+    // Simulate KCL v2 handing back a slice of a larger pooled buffer.
+    // .slice() produces a buffer with arrayOffset > 0; .array() on this returns
+    // the full backing array from index 0, which is the bug this test guards against.
```
domlander, line 31: What's a backing array?

domlander (approved):

test/modules/clustersync/TagEventDeserialiserTest.scala
Georges-GNM: AIUI, it basically just means the underlying data is stored in an array - see [ArrayList](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html) for example

AWS has some documentation [here](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/core/ResponseBytes.html#:~:text=Behavior%20by%20buffer%20type%3A,heap%20array.%20No%20concurrency%20risks.) which might be helpful too.

PR #631, opened 2026-04-08
Add deprecated field

### LINKED ISSUE #632: Add deprecated field to the backend
See [parent issue](https://github.com/guardian/tagmanager/issues/600) for more info

dblatcher:
## What does this change?

This PR adds a new field to the Tag Manager API responses (both the tag management API and the hyper API) for deprecating tags, i.e. indicating that they should no longer be used.

It does not add anything to the UI for now, so that we can merge these changes as-is and then implement functionality in Composer, Workflow, and CAPI to support the field, before finally allowing users to set it. (The PR does include [a commit that adds a checkbox to toggle the field](https://github.com/guardian/tagmanager/commit/4e99d74ec95ab957918ae5a551880734ae765ebc) and [a commit reverting that commit](https://github.com/guardian/tagmanager/commit/30efcd07de81fd6f91731c35229bed78cb8c600d), which we’ve left there so the latter can be re-reverted as a starting point when we’re ready to update the UI.)

To test this change we:

- ran tag manager locally
- opened an SSH tunnel to path-manager-CODE [as described in the readme](https://github.com/guardian/tagmanager/blob/30efcd07de81fd6f91731c35229bed78cb8c600d/readme.md#L38-L43)
- used the checkbox we added in [this commit](https://github.com/guardian/tagmanager/commit/4e99d74ec95ab957918ae5a551880734ae765ebc) to set the field on an existing tag
- confirmed that this was returned correctly from [the tag management API](https://tagmanager.local.dev-gutools.co.uk/api/tag/71301), from [the hyper API](https://tagmanager.local.dev-gutools.co.uk/hyper/tags/71301), and when [searching the hyper API](https://tagmanager.local.dev-gutools.co.uk/hyper/tags?query=+eric+rodriguez)
- created a new tag with the field set, confirmed that the set value was returned from the backend and visible in the frontend after refreshing the page

To test this yourself, you could either revert [our revert commit](https://github.com/guardian/tagmanager/commit/30efcd07de81fd6f91731c35229bed78cb8c600d) locally (to see the checkbox again), or you could use the API to create/edit tags (copying a browser request as a curl command is probably the easiest way to do this).


SHession (approved):
Not tested but the changes look sensible and carried out during pairing. Happy to rubberstamp

dblatcher:
double checked on CODE that the api response change didn't break Composer or Workflow

PR #648, opened 2026-05-15
Add video tracking subtype

waisingyiu:
## What does this change?

We are working on [making content atoms taggable](https://github.com/guardian/content-atom/pull/211) to improve data analysis.

The pull request adds a subtype `video` under `tracking` type for tags that will be intended for video tracking.

In tagmanager, create a new tag and on the UI we should be able to select the new subtype `video` under the `tracking` type. After creating a tag with video tracking subtype, we could find the tags by the API - e.g. https://tagmanager.code.dev-gutools.co.uk/hyper/tags?query=<tag name>&type=tracking&subType=video&limit=150

## Images

| Before | After |
| --- | --- |
| Screenshot 2026-05-15 at 15 33 55 | Screenshot 2026-05-15 at 15 40 47 |

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214714624861786

### DIFF
--- a/public/constants/trackingTagTypes.js
+++ b/public/constants/trackingTagTypes.js
@@ -1,5 +1,6 @@
 export const trackingTagTypes = [
   {name: 'Commissioning Desk', value: 'commissioningdesk'},
   {name: 'Platform Functional', value: 'platformfunctional'},
-  {name: 'Audience', value: 'audience'}
+  {name: 'Audience', value: 'audience'},
+  {name: 'Video', value: 'video'},
 ];


public/constants/trackingTagTypes.js
```diff
@@ -1,5 +1,6 @@
 export const trackingTagTypes = [
   {name: 'Commissioning Desk', value: 'commissioningdesk'},
   {name: 'Platform Functional', value: 'platformfunctional'},
-  {name: 'Audience', value: 'audience'}
+  {name: 'Audience', value: 'audience'},
+  {name: 'Video', value: 'video'},
```
SHession, line 5: AFAICT It doesn't look like this data is included in the [Tag Thrift definition](https://github.com/guardian/tags-thrift-schema/blob/main/src/main/thrift/tag.thrift). Is this the only place where the sub type is specified?

waisingyiu: Yes, I think so. The subtype is represented as a plain string in the Tag Thrift model - https://github.com/guardian/tags-thrift-schema/blob/main/src/main/thrift/tag.thrift#L129

SHession: I missed that!

SHession (approved):

PR #654, opened 2026-06-09
Use specific credentials provider for S3 so that local profiles are checked

davidfurey:
## What does this change?

This changes the provider chain used for S3.

I believe that running this project locally was broken by https://github.com/guardian/tagmanager/pull/638, and this PR fixes it by using a provider chain that tries the "composer" profile for S3 when running tagmanager locally.

### DIFF
--- a/app/services/AWS.scala
+++ b/app/services/AWS.scala
@@ -51,8 +51,14 @@ object AWS {
     .region(regionV2)
     .build()
 
+  lazy val credentialsProvider = DefaultCredentialsProvider
+    .builder()
+    .profileName("composer")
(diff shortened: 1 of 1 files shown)


Georges-GNM (approved):

davidfurey:
The project was NOT broken by that PR. Actually, the project only works if you run `./scripts/start.sh` which sets the environment variable `AWS_PROFILE` rather than running `sbt run` manually.

Since this PR only addresses the S3 clients reliance on `AWS_PROFILE` rather than covering all SDK clients, I'm going to close.

PR #669, opened 2026-09-17
Extend /api/tag/:id to support paths as well as IDs

liamness:
## What does this change?

If you pass a string to `/api/tag/:id` then it will now support accepting a string as the parameter, and will return a tag with a path matching this string if one is found.

This is motivated by the tag suggester functionality in Composer using `/hyper/tags?query=<path>`, when that endpoint is really intended for searching, not retrieving specific records. As such it returns tags that have paths which are only partial matches, meaning we have returned e.g. `uk/london-riots` when `uk/london` is what was asked for.

Will deploy to code and test in concert with the relevant changes in Composer.


liamness:
No longer needed: see #670

